PR #8, opened 2019-10-08
Models initial development

HeikoSchuett:
This is the basic development of the models module


HeikoSchuett:
Finally made it acceptable for code climate

pyrsa/model/__init__.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""Model definitions and handling
+"""
+from pyrsa.model.model import Model, ModelFixed
```
JasperVanDenBosch, line 5: we might find later that having a module with the same name as the package leads to a hassle. One option would be to name the classes and modules more specifically, i.e. `AbstractModel`, `GenericModel`. The module could then be `abstract_models` or something. But let's leave this for later :)

pyrsa/model/model.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    def __init__(self, name):
+        self.name = name
+        self.n_param = 0
+
+    def predict(self, theta=None):
+        """ Returns the predicted rdm(-vector)
+
+        Args:
+            theta(np.array 1d): the model parameter vector
```
JasperVanDenBosch: the precise name of the argument type here would be `numpy.ndarray`. If we use those, in the documentation it will generate a link to that class. The dimension expectations could then be in the description part

pyrsa/__init__.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+""" Top level package: Only imports and organisation
+"""
+import pyrsa.model as model
```
JasperVanDenBosch: I don't think this is necessary?

pyrsa/model/model.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class ModelFixed(Model):
+    """
+    Fixed model
+    This is a parameter-free model that simply predicts a fixed RDM
+    It takes a unidimension numpy-vector as an input to define the RDM
+    """
+    # Model Constructor
+    def __init__(self, name, rdm):
```
JasperVanDenBosch, line 54: So this would later accept a `pyrsa.Rdms` class instead of the numpy array?

pyrsa/__init__.py
HeikoSchuett: pylint complaint about it when it was missing on my computer. Would not think this is necessary either

pyrsa/model/model.py
HeikoSchuett: I think this is the plan

pyrsa/__init__.py
JasperVanDenBosch: It complained about a missing module? Did you try a `python setup.py develop`?

HeikoSchuett: I was talking about the comment. Are you talking about the import?

JasperVanDenBosch: yes

JasperVanDenBosch: tests still run if your remove it

HeikoSchuett: yes you are right -> removed

JasperVanDenBosch (approved):
good work! First real code

doerlbh:
looks good~

JasperVanDenBosch:
> looks good~

@doerlbh can you click `approve` under the `review` menu?

doerlbh (approved):
looks good

PR #16, opened 2019-10-15
Data dev

jdiedrichsen:
Data set class definition - 
I know that we talked about having the data set only containing a 2d-array
However, considering how we designed the RDMs class, I think it would be consistent to allow also 3d-arrays - with the variable n_set containign the number of data sets 
This makes calculations on data sets with the same observation structure less awkward and generalizes nices. 
obs_descriptors and channel_descriptors are forced to be the same over all sets - otherwise you need to make a new Dataset object 

We may want an open discussion before this is merged into master

Indicator module in util 
This is a collection of useful indicator matrices for coding linear models, pairwise contrast, etc.


doerlbh (approved):
I agree with you that having a 3d Dataset can make it more flexible. I will integrate my updates in your initialized version and pushed it tonight.

doerlbh:
Hi guys, I also added the options where people can implement their own preprocessing function for arbitrary data format (as option args _rawdata_ and _preprocess_) - they can be found useful in files neurodataset.py and exampledataset.py where submodules can be implemented there.

doerlbh:
@jdiedrichsen Regarding the 3d-arrays, with the variable n_set containing the number of data sets, I now have a different opinion:

I like the additional flexibility of mulitiple sets of datasets. However, I believe this might create much additional burdens to the data structures of _descriptor_,  _obs_descriptor_, and _channel_descriptor_ - are they also need to have an additional dimension for the n_set?

An alternative would be, as we also briefly chatted on Oct 7, is to have an additional class, called _DatasetList_ or _DatasetSequence_, where the Dataset object are collated together in a list. This also enables additional integration of my part on representational dynamics analysis, where RDM movies will be required to create from sliding window data, in this case, a DatasetList/DatasetSequence.

What do you think?

Other than this, I believe the pull request is still valid, since the 3d/2d discussion doesn't interact with other modules in this current commits (which should have already accommodated for all 2d inputs, with 3d inputs as an option).

doerlbh:
All four methods have been implemented for Dataset class, accommodating 2d measurements as input:

**split_obs**(by=’descriptor’) returns list of Datasets
**split_channel**(by=’descriptor’) returns list of Datasets
**subset_obs**(by=‘descriptor’,value=’value’) returns Dataset
**subset_channel**(by=‘descriptor’,value=’value’) returns Dataset

JasperVanDenBosch:
> An alternative would be, as we also briefly chatted on Oct 7, is to have an additional class, called _DatasetList_ or _DatasetSequence_, where the Dataset object are collated together in a list. 

Agreed

JasperVanDenBosch:
Can either of you have a look at the style issues on codeclimate? See the *details* link below next to CodeClimate. Many of them are just about whitespace. If you have a linter extension in your IDE it will mark these for you with squiggly lines, or alternatively you can automatically fix these with something like `autopep8`. Some other issues are repetitions of code, which can be fixed with a helper method.

JasperVanDenBosch:
Also the unit tests are currently failing, see Travis:

```
  File "/home/travis/build/rsagroup/pyrsa/pyrsa/data/dataset.py", line 40, in DatasetBase
    def split_obs(self, by=descriptor):
NameError: name 'descriptor' is not defined
```

HeikoSchuett:
I fixed the test problem in the rdm-dev branch. This should give an idea how to fix it here: essentially remove discriminator from by= discriminator as discriminator is not a defined variable

doerlbh:
Thanks for the pointers on the unit tests and style issues! Now the unit test issues are all fixed and committed. And will work on the style in pyrsa.data later this week as well.

doerlbh:
@jdiedrichsen @ilogue @HeikoSchuett @iancharest Feel free to checkout the latest version of **data-dev**. All tests are passing (unit tests and style tests). It would be great to merge the current stage soon, before more developmental changes are made.

pyrsa/data/dataset.py
```diff
@@ -0,0 +1,165 @@
...
+            by(String): the descriptor by which the subset selection
+                        is made from obs dimension
+            value:      the value by which the subset selection is made
+                        from obs dimension
+
+        Returns:
+            Dataset, with subset defined by the selected obs_descriptor
+        """
+        return self.measurements[:, self.obs_descriptors[by] == value, :]
+        # TODO: for 3d measurements, need implementations.
```
JasperVanDenBosch: please convert the *TODO* comments to an issue before merging. I think the master branch should not contain *TODOs*.

pyrsa/data/dataset.py
```diff
@@ -0,0 +1,165 @@
...
+            by(String): the descriptor by which the splitting is made
+
+        Returns:
+            list of Datasets, splitted by the selected obs_descriptor
+        """
+        unique_values = set(self.obs_descriptors[by])
+        dataset_list = []
+        for v in unique_values:
+            dataset_list.append(
+                self.measurements[:, self.obs_descriptors[by] == v, :])
```
JasperVanDenBosch: For future reference, another way to create shorter lines, is to break up the statement by creating a separate (local scope) indexing variable. For instance:

```
selection = self.obs_descriptors[by] == v, :]
dataset_list.append(self.measurements[:, selection])

```

pyrsa/util/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+utility functions
+"""
+
+# from pyrsa.util.indicator import identity, allpairs, identity_pos
```
JasperVanDenBosch: perhaps remove commented-out code before merging to master

tests/test_data.py
```diff
@@ -0,0 +1,30 @@
...
+test_data
+Test for Dataset class 
+@author: jdiedrichsen
+"""
+
+import unittest 
+import pyrsa.data as rsd
+import numpy as np 
+
+class TestSimulation(unittest.TestCase):
```
JasperVanDenBosch: See my comments in #17, should this be named `TestSimulation`?

JasperVanDenBosch (changes requested):

tests/test_data.py
doerlbh: Will work on this next week, and add more test cases. Thanks for the note!

pyrsa/data/dataset.py
doerlbh: Thanks for the note! Will clean up and push next week.

pyrsa/data/dataset.py
doerlbh: Good idea! Will change similar structure and push next week.

pyrsa/util/__init__.py
doerlbh: Thanks for the note! Will clean up and push next week.

HeikoSchuett:
For all functions of the dataset (split, subset etc.) you currently only return the measurements. Shouldn't this be new dataset objects with descriptors attached?

pyrsa/data/exampledataset.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
```
HeikoSchuett: How is this going to be different from Dataset/DatasetBase? Currently it looks like a duplicate

pyrsa/data/dataset.py
```diff
@@ -0,0 +1,165 @@
...
+                                          Dataset object
+            obs_descriptors (dict):       observation descriptors (all
+                                          are array-like with shape =
+                                          (n_obs,...))
+            channel_descriptors (dict):   channel descriptors (all are
+                                          array-like with shape =
+                                          (n_channel,...))
+        Returns:
+            dataset object
+    """
```
HeikoSchuett, line 34: Maybe the init for the dataset objects should do some checks that the descriptors are all compatible with the size of the measurement matrix?

doerlbh:
> For all functions of the dataset (split, subset etc.) you currently only return the measurements. Shouldn't this be new dataset objects with descriptors attached?

Good point. Will change to this in the next few pushes. Thanks!

pyrsa/data/dataset.py
doerlbh: Will include this in the next few pushes. Thanks!

pyrsa/data/exampledataset.py
doerlbh: The Dataset class is the default/vanilla DatasetBase child that takes in a standard input and perform standard operations. The DatasetBase parent enables customized inputs and operations, as in exampledataset.py and neurodataset.py. The user should be able to simply use Dataset for their standard input without further coding, while more advanced user might wish to build upon the DatasetBase for their customized data operations.

HeikoSchuett: Yes, that is all clear, but what will exampledataset do? 
Is this an example how a user would define their own dataset? Or what is this specific file for?

doerlbh: Yes, it is an example how a user would define their own dataset. For users that are not expert programmers, it would be good to have an example for them to build on, where they can just implement the part of preprocess function.

pyrsa/data/dataset.py
doerlbh: cleaned up and pushed.

pyrsa/data/dataset.py
doerlbh: fixed and pushed.

pyrsa/data/dataset.py
doerlbh: > For all functions of the dataset (split, subset etc.) you currently only return the measurements. Shouldn't this be new dataset objects with descriptors attached?

fixed and pushed.

doerlbh:
> For all functions of the dataset (split, subset etc.) you currently only return the measurements. Shouldn't this be new dataset objects with descriptors attached?

fixed and pushed.

pyrsa/util/__init__.py
doerlbh: fixed and pushed.

pyrsa/data/dataset.py
doerlbh: fixed and pushed.

tests/test_data.py
doerlbh: fixed and pushed. As well as including test cases for initialization and four methods, all past unit tests.

doerlbh:
Implementations and unit tests for the class with four main functions are all now complete and passing all criteria. Helpful remarks and requested changes from @ilogue and @HeikoSchuett are all fixed in the current version. Please review the code and let me know if anything else need to be changed, and approve if you think it's ready. Thanks!

doerlbh:
Implementations and unit tests for the class with two main functions are all now complete and passing all criteria. Helpful remarks and requested changes from @ilogue are all fixed in the current version. Please review the code and let me know if anything else need to be changed, and approve if you think it's ready. Thanks!

JasperVanDenBosch (approved):
Happy with this!

PR #17, opened 2019-10-15
Rdm dev

doerlbh:
Hi guys, I created a vanilla case for RDM class for people who is eager to use. The methods are still pending to be further implemented though, including transforms and dissimilarity computation etc..


doerlbh:
Added RDMs class and implemented the two main functions:
**get_vectors**()
**get_matrices**()

HeikoSchuett:
I shortly went and fixed the unit-test issues. Please have a look through the code climate report and fix those. You can also run these tests with pylint pyrsa locally. 
Otherwise: Great that we now have something to work on.

pyrsa/util/indicator.py
```diff
@@ -0,0 +1,62 @@
...
+    """ Indicator matrix with one column per unique positive element in vector  
+        Args:
+            c (numpy.ndarray): n_row vector to code - discrete values (one dimensional)
+        Returns:
+            X (numpy.ndarray): n_row x n_values indicator matrix 
+    """
+    cc = np.unique(c)
+    K = cc.size
+    rows = np.size(c)
+    cc = cc[cc>0]
```
JasperVanDenBosch: On first look it seems like `identity` and `identity_pos` are quite similar, perhaps  `identity_pos` could call `identity` with an flag `positive=True`? It then checks for this flag and does the `cc[cc>0]` step. This solves the code replication warning.

doerlbh:
Will work on the code climate stuff later this week or early next week. Other than that, the unit tests now are all passing with the latest updates from data-dev.

doerlbh:
@jdiedrichsen @ilogue @HeikoSchuett @iancharest Feel free to checkout the latest version of **rdm-dev**. All tests are passing (unit tests and style tests). It would be great to merge the current stage soon, before more developmental changes are made.

pyrsa/data/dataset.py
```diff
@@ -0,0 +1,165 @@
...
+"""
+Definition of RSA Dataset class and subclasses
+@author: baihan, jdiedrichsen
+"""
+
+import numpy as np
+import pyrsa as rsa
+
+
+class DatasetBase:
```
JasperVanDenBosch: I think these should go in #16 right?

tests/test_data.py
```diff
@@ -0,0 +1,30 @@
...
+test_data
+Test for Dataset class 
+@author: jdiedrichsen
+"""
+
+import unittest 
+import pyrsa.data as rsd
+import numpy as np 
+
+class TestSimulation(unittest.TestCase):
```
JasperVanDenBosch: Is this for the Simulation package? It does not seem to test RDM code?

pyrsa/rdm/__init__.py
```diff
@@ -0,0 +1 @@
+from pyrsa.data.rdm import RDMs
```
JasperVanDenBosch: I think this should be `from pyrsa.rdm.rdms import RDMs`

pyrsa/util/indicator.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    c_unique = np.unique(index_vector)
+    n_unique = c_unique.size
+    rows = np.size(index_vector)
+    c_unique = c_unique[c_unique > 0]
+    n_unique = c_unique.size
+    indicator_matrix = np.zeros((rows, n_unique))
+    for i in range(n_unique):
+        indicator_matrix[index_vector == c_unique[i], i] = 1
+    return indicator_matrix
```
JasperVanDenBosch: On first look it seems like `identity` and `identity_pos` are quite similar, perhaps  `identity_pos` could call `identity` with an flag `positive=True`? It then checks for this flag and does the `cc[cc>0]` step. This solves the code replication warning.

tests/test_indicator.py
```diff
@@ -0,0 +1,39 @@
...
+        n_row,n_col = X.shape
+        self.assertEqual(n_row,10)
+        self.assertEqual(n_col,4)
+
+    def test_allpairs(self):
+        a = np.array(range(0,5))
+        X = rsu.indicator.allpairs(a)
+        n_row,n_col = X.shape
+        self.assertEqual(n_row,10)
+        self.assertEqual(n_col,5)
```
JasperVanDenBosch: In the future, it may be useful also to test the contents, e.g. by testing one specific cell. Unless only the shapes of these matrixes matter? I am not very familiar with the concept of indicator matrices.

JasperVanDenBosch (changes requested):
Good work on the style fixes, I have a few more requests, in particular which PR will contain the dataset classes and to reduce the complexity of the indicator module.

pyrsa/data/dataset.py
doerlbh: Yes, that's true, but since I accidentally start my branch rdm-dev from the data-dev, so I have to manually update it in rdm-dev from data-dev, before pull #16 is approved and merged.

tests/test_data.py
doerlbh: I haven't implemented the test part yet for the rdm classes. I plan to do that as a next step (perhaps in the next pull), or do you suggest me to provide the test now in this pull?

pyrsa/rdm/__init__.py
doerlbh: fixed and pushed.

tests/test_data.py
doerlbh: I will add test_rdm.py in my next few pushes. Thanks for the note.

pyrsa/util/indicator.py
doerlbh: Thanks for the note! Will take a look and change in the next few pushes.

pyrsa/util/indicator.py
doerlbh: Thanks for the note! Will take a look and change in the next few pushes.

tests/test_indicator.py
doerlbh: Thanks for the note! Will take a look and change in the next few pushes. I think the shape is important to check to avoid dimension conflicts, while adding content check will definitely be essential as well.

pyrsa/data/dataset.py
JasperVanDenBosch: I'm sorry I don't think I understand. Why's that? Are the dataset classes imported by the rdm modules?

tests/test_data.py
doerlbh: fixed and pushed.

pyrsa/util/indicator.py
doerlbh: fixed and pushed.

pyrsa/util/indicator.py
doerlbh: fixed and pushed.

tests/test_indicator.py
doerlbh: Fixed and pushed. As suggested, I added a simple content check. For more extensive content checks, I think Joern would be a better person to add them as he first implemented indicator functions and is more familiar to the concept, and might know which cells are more important to add in test cases.

doerlbh:
Implementations and unit tests for the class with two main functions are all now complete and passing all criteria. Helpful remarks and requested changes from @ilogue are all fixed in the current version. Please review the code and let me know if anything else need to be changed, and approve if you think it's ready. Thanks!

JasperVanDenBosch:
@doerlbh The next step on this I think is to merge master (remove the Data contents from this branch)

doerlbh:
@ilogue Thanks! FYI, I have merged all the updates from the master branch, and they are all cleaned up. Feel free to check it out~

JasperVanDenBosch (approved):
Looks good!

saarbuckle (approved):
Good stuff

PR #18, opened 2019-10-15
Vis dev

doerlbh:
Initialized vis modules, added modified MDS as reference.


iancharest:
Hi Baihan, 
out of curiosity, for the MDS, could we not just wrap our class / def around sklearn manifold?
```python 
from sklearn import manifold
X = RDM
mds = manifold.MDS(2, max_iter=10000, n_init=1,
                   random_state=5, dissimilarity='precomputed')
Y = mds.fit_transform(X)
```

doerlbh:
Hi Ian,

Thanks for the comment and suggestion! In fact, the reason why I am including a customized version here is for the purpose of my integration of "**weighted**" MDS - the standard MDS from sklearn.manifold doesn't have that option or flexibility in their source code. Other than this feature, the customized code so far is mostly the same from the standard sklearn.manifold.mds.

Best,
Baihan

JasperVanDenBosch:
@doerlbh one more thing regarding the MDS - when copying code from another source we must make sure to follow the proper legal procedure, in this case to include the license. Perhaps we can put the BSD `LICENSE.txt` in the `mds.py` module? Or as a file in the `vis` package? This is just to make sure we won't get sued in the future

doerlbh:
Hi Jasper, thanks for the pointers on the license issue. I will find the original code and include the license later this week. I just put that on the todo list as well.

Will work on the code climate stuff later this week or early next week. Other than that, the unit tests now are all passing with the latest updates from data-dev.

iancharest:
Hi @doerlbh can you check the The Travis CI build fails on this branch?

doerlbh:
> Hi @doerlbh can you check the The Travis CI build fails on this branch?

Onto it. Thanks for the note.

doerlbh:
> Hi @doerlbh can you check the The Travis CI build fails on this branch?

fixed.

doerlbh:
Questions: other than the dimension reduction such as MDS, what other visualization modules should we include for the basic RSA (as discussed earlier, dynamics etc. should be in the next iteration)? Is the dimension reduction transform itself sufficient? or do we also want a basic plot generating function, say, a python code snippet to output a png file of MDS? What do you all think?

Either way, the current latest commit has a basic MDS transform with a test, given a RDMs object, if you wish to review and merge.

JasperVanDenBosch:
> Questions: other than the dimension reduction such as MDS, what other visualization modules should we include for the basic RSA (as discussed earlier, dynamics etc. should be in the next iteration)? Is the dimension reduction transform itself sufficient? or do we also want a basic plot generating function, say, a python code snippet to output a png file of MDS? What do you all think?


There are some more todo's on the project page but I think this is enough for an MDS PR.

JasperVanDenBosch (approved):
thanks @doerlbh, I've made some small style fixes but otherwise this looks good to me.

HeikoSchuett (approved):
This looks fine to me. I think ultimately we do want to have MDS plotting utility, not only calculating it, but the code here looks fine for the MDS itself

doerlbh:
Now we have enough approvals. Ready to merge (by authorized users).

Issue #19, opened 2019-10-17
RDM colormap


JasperVanDenBosch:
can import from Charestlab/pyrsa

doerlbh:
I think this is a great idea.

iancharest:
fixed with 6ebc632

PR #21, opened 2019-10-19
added colors.py [color_scale, rdm_colormap] and test_colors.py [passed]

iancharest:
unit tests have passed on my machine. please review.

### DIFF
--- a/tests/test_colors.py
+++ b/tests/test_colors.py
@@ -23,8 +23,7 @@ class ColorTests(unittest.TestCase):
         n_cols = 10
         cols = rdm_colormap(n_cols)
         n_cols_returned, n_rgb = cols.colors.shape
+        last_color = [1., 1., 0]
         self.assertEqual(n_cols_returned, n_cols)
         self.assertEqual(n_rgb, 3)
-
-if __name__ == '__main__':
-    unittest.main()
+        self.assertEqual(last_color, cols.colors[-1])
--- a/tests/test_colors.py
+++ b/tests/test_colors.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+test_colors
+Test for colors 
+@author: iancharest
+"""
+
+import unittest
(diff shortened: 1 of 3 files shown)


tests/test_colors.py
```diff
@@ -0,0 +1,30 @@
...
+test_colors
+Test for colors 
+@author: iancharest
+"""
+
+import unittest
+from pyrsa.vis.colors import color_scale
+from pyrsa.vis.colors import rdm_colormap
+
+class ColorTests(unittest.TestCase):
```
JasperVanDenBosch, line 14: Perhaps you can add a quick test that actually tests the colors, with a dummy rdm for example, checking that cell 1 is brighter than cell2 or whatever the goal of the colormap is. Currently if the function just returns a bunch of nans, you'd never know

tests/test_colors.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_rdm_colormap(self):
+        n_cols = 10
+        cols = rdm_colormap(n_cols)
+        n_cols_returned, n_rgb = cols.colors.shape
+        self.assertEqual(n_cols_returned, n_cols)
+        self.assertEqual(n_rgb, 3)
+
+if __name__ == '__main__':
+    unittest.main()
```
JasperVanDenBosch: this part is not necessary I think, if you want to run one test case you can call:

 `python setup.py test -s tests.test_colors.ColorTests`

JasperVanDenBosch (changes requested):

JasperVanDenBosch:
resolves #19

tests/test_colors.py
iancharest: working on it.

tests/test_colors.py
iancharest: i've added a test case. the last colour returned from rdm_colormap should by design be yellow.

JasperVanDenBosch (approved):

doerlbh (approved):

Issue #25, opened 2019-11-04
Implement basic RDM calculations from dataset


HeikoSchuett:
At least:

- Euclidean distance
- Correlation distance
- Mahalanobis distance
- Crossnobis distance

PR #27, opened 2019-11-04
Calculating RDMs

HeikoSchuett:
This will be the code to calculate RDMs from datasets


HeikoSchuett:
I'll have a look at the refactoring suggestions of codeclimate and implement the missing cosine distance.
Then the only remaining issue is the stability of the crossnobis ram to imbalanced designs

jdiedrichsen (commented):
@ Heiko: The functionality of providing a contrast matrix is already done in util.indicator.allpairs 
(all pairwise contrasts). 
It's slightly more general, but util.indicator.allpairs(range(0,n_patterns)) does the same job, I think
Is there a reason why we need this extra version?

HeikoSchuett:
No, there is no reason to double it. I just removed it. I was simply not aware of all pairs existence when I copied the code from my code here

HeikoSchuett:
I would now suggest to merge this.
This does not do all variants of RDMs we might want, but we can add more later and I have issues on them open already. 

So: Please review and raise points where I need to change things still

HeikoSchuett:
closes #25

pyrsa/rdm/calc_rdm.py
```diff
@@ -0,0 +1,200 @@
...
+    they are averaged.
+
+        Args:
+            dataset (pyrsa.data.DatasetBase):
+                The dataset the RDM is computed from
+            descriptor (String):
+                obs_descriptor used to define the rows/columns of the RDM
+                defaults to one row/column per row in the dataset
+            noise (numpy.ndarray):
+                precision matrix used to calculate the RDM
```
JasperVanDenBosch: Could you add something about the expected dimensions of the noise array?

JasperVanDenBosch (changes requested):

pyrsa/rdm/calc_rdm.py
HeikoSchuett: I added a line to the help that this is supposed to have dimension n_channel x n_channel and introduced corresponding checks in a new commit

JasperVanDenBosch:
unit test in progress.. will commit tomorrow

JasperVanDenBosch:
@HeikoSchuett the import `from pyrsa.rdm.calc_rdm import calc_rdm` in `pyrsa.rdm.__init__.py` currently 'hides' the `calc_rdm` module (as they have the same name). This means it's not possible to import he module itself anymore I think. Should we rename the module? Or the import shortcut?

HeikoSchuett:
> @HeikoSchuett the import `from pyrsa.rdm.calc_rdm import calc_rdm` in `pyrsa.rdm.__init__.py` currently 'hides' the `calc_rdm` module (as they have the same name). This means it's not possible to import he module itself anymore I think. Should we rename the module? Or the import shortcut?

To fix this issue I now renamed the submodule into `calc`. Thank you for finding this!

I also changed the test you wrote to compare to the squared Euclidean distance divided by the number of channels as this is what the calc_rdm_euclid computes. It now passes.

JasperVanDenBosch:
Great. I think this could serve as a good example of how to make a unit test that actually tests the main objective of the unit. Also it only tests the specific unit by mocking out the other class involved. (If dataset has a bug, it should not be the rdm tests that fail). The main benefit of this is that it shows the dependencies between the two classes clearly.

I'm ok with merging this now and then adding more tests latsr

JasperVanDenBosch (approved):

g14r (approved):

PR #32, opened 2019-11-05
add output functions for objects

HeikoSchuett:
I think it is very helpful to define the __str__ and __repr__ functions for our objects to give a more helpful output when you type dataset or rdm in your terminal. 

Here I will implement these for dataset & RDM. If you think other information should be displayed or a different format should be used please say so!

### DIFF
--- a/pyrsa/data/dataset.py
+++ b/pyrsa/data/dataset.py
@@ -33,8 +33,8 @@ class DatasetBase:
         Returns:
             dataset object
     """
-    def __init__(self, measurements, descriptors={},
-                 obs_descriptors={}, channel_descriptors={}):
+    def __init__(self, measurements, descriptors=None,
+                 obs_descriptors=None, channel_descriptors=None):
         if measurements is not None:
             if measurements.ndim != 2:
                 raise AttributeError(
@@ -46,9 +46,18 @@ class DatasetBase:
                                     self.n_obs)
         check_descriptors_dimension(channel_descriptors, "channel_descriptors",
                                     self.n_channel)
-        self.descriptors = descriptors
-        self.obs_descriptors = obs_descriptors
-        self.channel_descriptors = channel_descriptors
+        if descriptors is None:
+            self.descriptors = {}
+        else:
+            self.descriptors = descriptors
+        if descriptors is None:
+            self.obs_descriptors = {}
+        else:
+            self.obs_descriptors = obs_descriptors
+        if descriptors is None:
+            self.channel_descriptors = {}
+        else:
+            self.channel_descriptors = channel_descriptors
 
     def __repr__(self):
         """
--- a/pyrsa/data/dataset.py
+++ b/pyrsa/data/dataset.py
@@ -10,6 +10,7 @@ import pyrsa as rsa
 from pyrsa.util.data_utils import check_descriptors_dimension
 from pyrsa.util.data_utils import extract_dict
 from pyrsa.util.data_utils import get_unique_unsorted
+from pyrsa.util.descriptor_utils import format_descriptor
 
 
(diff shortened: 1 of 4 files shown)


pyrsa/data/dataset.py
```diff
@@ -32,8 +33,8 @@ class DatasetBase:
         Returns:
             dataset object
     """
-    def __init__(self, measurements=None, descriptors=None,
-                 obs_descriptors=None, channel_descriptors=None):
+    def __init__(self, measurements, descriptors={},
+                 obs_descriptors={}, channel_descriptors={}):
```
JasperVanDenBosch: Using a mutable object like a `dict` or `list` for an argument default value is generally a bad idea, as the changed object will later server as the default again..

pyrsa/data/dataset.py
```diff
@@ -49,6 +50,35 @@ def __init__(self, measurements=None, descriptors=None,
...
+        """
+        defines string which is printed for the object
+        """
+        return (f'pyrsa.data.{self.__class__.__name__}(\n'
+                f'measurements = \n{self.measurements}\n'
+                f'descriptors = \n{self.descriptors}\n'
+                f'obs_descriptors = \n{self.obs_descriptors}\n'
+                f'channel_descriptors = \n{self.channel_descriptors}\n'
+                )
+
```
JasperVanDenBosch, line 62: This works, another approach would be a triple-quoted f string:

```
f"""
pyrsa.data.{self.__class__.__name__}(
    measurements = {self.measurements}
    descriptors = {self.descriptors}
    obs_descriptors = {self.obs_descriptors}
    channel_descriptors = {self.channel_descriptors}
)
"""
```

JasperVanDenBosch (changes requested):
Good work, this will definitely make it more user friendly. I'd recommend against the mutable defaults, otherwise looks good. I'll contribute a test

pyrsa/util/descriptor_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+        Returns:
+            string_descriptors(String): formated string to show dict
+    """
+    string_descriptors = ''
+    for entry in descriptors:
+        string_descriptors = (string_descriptors +
+                              f'{entry} = {descriptors[entry]}\n'
+                              )
+    return string_descriptors
```
JasperVanDenBosch, line 23: how about `return ''.join([f'{k} = {v}\n' for k, v in descriptors.items()])`? Too terse perhaps?

pyrsa/data/dataset.py
HeikoSchuett: That is definitely good to know. I did not expect that behavior here. Fixed in a new commit.

JasperVanDenBosch (approved):

pyrsa/util/descriptor_utils.py
HeikoSchuett: well, is it faster or in some other way better than the loop? If so, I think it is hidden enough that we could change to your one-line solution; it is hidden enough after all.

pyrsa/data/dataset.py
HeikoSchuett: True, in this case I like the one string per output version. Do you have a strong preference for the triple-quoted string?

pyrsa/util/descriptor_utils.py
JasperVanDenBosch: No both are fine

pyrsa/data/dataset.py
JasperVanDenBosch: nope, no preference

g14r (approved):

PR #34, opened 2019-11-07
util.indicator renamed into util.matrix

jdiedrichsen:
Made the naming of utility matrices more intuitive. 
Added centering matrix
util.matrix.indicator 
util.matrix.pairwise_contrast
util.matrix.centering


saarbuckle (approved):
Looks good

HeikoSchuett (approved):
I now also moved the descriptor handling into the appropriate file.
please check again.
This will break some stuff in ongoing development branches! Check there once we merge.

iancharest (changes requested):
This looks good but I think it wouldn't hurt to be more explicit for the indicator docs. This is a very useful tool and with better documentation this will help people realise that they can compute distances using mat multiplication.

HeikoSchuett:
> This looks good but I think it wouldn't hurt to be more explicit for the indicator docs. This is a very useful tool and with better documentation this will help people realise that they can compute distances using mat multiplication.

should we maybe put this into a documentation somewhere? I'm not really sure how to put this into the help messages for the individual matricies. 

Also It would be convenient to merge this into master such that we can use it for development everywhere.

iancharest (approved):

Issue #39, opened 2019-11-13
alternative to np.corrcoef for correlation distance


iancharest:
Hi @HeikoSchuett ,

given that we use einsum for all other RDM calc cases, I figured we could also use it for correlation distance. that speeds up performance massively compared to numpy corrcoef.

```python
def correlation_distance(X):
    """[correlation distance]

    Args:
        X ([dataset]): conditions x voxels dataset

    Returns:
        RDM in upper triangular form.        
    """
    X = X - X.mean(axis=1, keepdims=True)
    X /= np.sqrt(np.einsum('ij,ij->i', X, X))[:, None]

    m = X.shape[0]
    r, c = np.triu_indices(m, 1)
     
    return 1 - np.einsum('ik,jk', X, X))[r, c]
```

HeikoSchuett:
Hi Ian,
sounds very sensible. We could then use the numpy implementation as the test in parallel to the euclidean implementation. 
Could you start this as a pull request with the new implementation of the correlation distance build in?

PR #41, opened 2019-11-25
corr dist using einsum and test case using np.corrcoef

iancharest:
Hi all,
I've implemented corr dist using einsum for symmetry with the other distance metrics. 
on my machine it passes the tests. but i am relatively new to einsum so a second opinion would be appreciated.


HeikoSchuett (approved):
Looks fine to me! Sorry for the wait. I was away last week. 

Did you check whether it is actually faster this way?

HeikoSchuett:
closes #39

eberlot (approved):
This implementation also looks fine to me.

doerlbh (approved):
Looks good to me. Thanks!

Issue #26, opened 2019-11-04
Implement distances between RDMs


HeikoSchuett:
At least:
- correlation distance
- rank correlation distance
- Kendal-Tau
- cosine distance

PR #42, opened 2019-12-06
differences between rdms

HeikoSchuett:
This will be the implementation of different difference metrics for RDMs


HeikoSchuett:
closes #26

pyrsa/rdm/compare.py
```diff
@@ -0,0 +1,178 @@
...
+            vectors1 (numpy.ndarray):
+                second set of values
+            func (function):
+                function to be applied, should take two input vectors
+                and return one scalar
+        Returns:
+            value (float):
+                average function result over all pairs
+    """
+    sum_val = 0
```
JasperVanDenBosch: any reason not to `numpy.apply_along_axis(func, -1, numpy.concatenate(vectors1, vectors2)).mean(axis=-1)` or something?

JasperVanDenBosch (changes requested):
Aside from `_average_all_combinations` perhaps being faster with numpy, looks good

pyrsa/rdm/compare.py
HeikoSchuett: well, essentially only that I did not find a function in numpy which does what I want. I think yours would assume that rdm1 and rdm2 have equally many entries and then only compare the matching ones? I also thought that there should be some kind of outer product implementation which works like this function, but I was unable to locate it. If anyone knows, please tell me and I will replace this.

JasperVanDenBosch: Ah yes I assumed they would be the same size. I didn't realize you could compare RDMs of different sizes.

HeikoSchuett: well, different numbers of RDMs not different sizes of RDMs in this case

HeikoSchuett: Anyway: If anyone out there knows how to do this with numpy/scipy, please tell me, otherwise I'll assume this is resolved once another review comes in.

JasperVanDenBosch: But if it's about missing data don't you want to just output NaNs where appropriate?

HeikoSchuett: Hmmm yes, but that is not the case here? The aim here is: I have two sets of RDMs: For example from different subjects and from different instances of a model. I cannot simply average them, because they may have different scaling, which I want to ignore. To solve this, I correlate each RDM from group 1 with each RDM from group 2 and average the results. 

This is more akin to a generalized outer product + then average than to applying along a dimension & I did in fact not care about missing values yet.

JasperVanDenBosch: Ah..ok I think you lost me :) but code wise :+1:

iancharest: Hi @HeikoSchuett i'm not sure I follow what you are trying to do. So if you have 2 sets of RDMs. say one set where you have 3x2556 (3 regions of interest, 3 subjects, whatever..) and you have 8 models 8x2556, you want to get back a 3x8 of correlations?

iancharest: Would a function like this one help?
```python
def corr_rdms(X, Y):
    """ corr_rdms correlate two stacks of RDMs.
        
    Args:
        X [array]: e.g. a stack of fmri ROI RDMs (n_rois x n_pairs)
        Y [array]: e.g. a stack of model RDMs (n_layers x n_pairs)

    Returns:
        [array]: correlations between X and Y, of shape dim0 of X by dim0 of Y
    """
    X = X - X.mean(axis=1, keepdims=True)
    X /= np.sqrt(np.einsum('ij,ij->i', X, X))[:, None]
    Y = Y - Y.mean(axis=1, keepdims=True)
    Y /= np.sqrt(np.einsum('ij,ij->i', Y, Y))[:, None]

    return np.einsum('ik,jk', X, Y)
```

JasperVanDenBosch (approved):

pyrsa/rdm/compare.py
HeikoSchuett: Hi @iancharest, 
here I immediately averaged to get a single value for the comparison of two rdms objects. Your comments got me thinking that your version and averaging if we need to elsewhere might be a better idea.  

I definitely like your corr_rdms function. I'll go and switch to that way of calculating for cosine and correlation distance.

HeikoSchuett: Hi @iancharest,
changes are now implemented. Does his seem right to you?

iancharest (changes requested):
Hi Heiko, this looks good. 

1 - I would add some notes to the documentation about kendall. This is kendall tau b that's implemented in scipy. @lajnd has implemented a kendall tau a based on scipy which is super fast, perhaps Daniel, you could include this here?
2 - Also, perhaps for the Spearman R, we could instead use a ranktransform before computing pearson corr? 
```python
import numpy
from scipy.stats import rankdata
def compare_spearman(rdm1, rdm2):
    """
    calculates the correlation distance between two RDMs objects

        Args:
            rdm1 (pyrsa.rdm.RDMs):
                first set of RDMs
            rdm2 (pyrsa.rdm.RDMs):
                second set of RDMs
        Returns:
            dist (float):
                correlation distance between the two RDMs
    """
    vector1, vector2 = _parse_input_rdms(rdm1, rdm2)
    vector1 = np.apply_along_axis(rankdata,1,vector1)
    vector2 = np.apply_along_axis(rankdata,1,vector2)
    vector1 = vector1 - np.mean(vector1, 1, keepdims=True)
    vector2 = vector2 - np.mean(vector2, 1, keepdims=True)
    sim = _cosine(vector1, vector2)
    return 1 - sim
```

notice also in the above how I changed the return from _cosine from "dist" to "sim" for clarity? because Return 1 - dist makes it sound like we are returning the similarity? This last comment applies to all functions in rdm.compare

HeikoSchuett:
> Hi Heiko, this looks good.
> 
> 1 - I would add some notes to the documentation about kendall. This is kendall tau b that's implemented in scipy. @lajnd has implemented a kendall tau a based on scipy which is super fast, perhaps Daniel, you could include this here?
> 2 - Also, perhaps for the Spearman R, we could instead use a ranktransform before computing pearson corr?
> 
> ```python
> import numpy
> from scipy.stats import rankdata
> def compare_spearman(rdm1, rdm2):
>     """
>     calculates the correlation distance between two RDMs objects
> 
>         Args:
>             rdm1 (pyrsa.rdm.RDMs):
>                 first set of RDMs
>             rdm2 (pyrsa.rdm.RDMs):
>                 second set of RDMs
>         Returns:
>             dist (float):
>                 correlation distance between the two RDMs
>     """
>     vector1, vector2 = _parse_input_rdms(rdm1, rdm2)
>     vector1 = np.apply_along_axis(rankdata,1,vector1)
>     vector2 = np.apply_along_axis(rankdata,1,vector2)
>     vector1 = vector1 - np.mean(vector1, 1, keepdims=True)
>     vector2 = vector2 - np.mean(vector2, 1, keepdims=True)
>     sim = _cosine(vector1, vector2)
>     return 1 - sim
> ```
> 
> notice also in the above how I changed the return from _cosine from "dist" to "sim" for clarity? because Return 1 - dist makes it sound like we are returning the similarity? This last comment applies to all functions in rdm.compare

I think I implemented those recommendations now.

If you, @lajnd, want to contribute your code here, this is very welcome. Simply implement it as another compare_kendall_tau_a or something similar and push it to this branch. 

Otherwise, I would hope to merge this soon. If it is already merged Daniel can still put it a new pull request to add the fast Kendall-tau A

iancharest (approved):
This looks good. we'll look into integrating kendall tau a and b..

PR #45, opened 2019-12-09
subset functions for RDMs objects

HeikoSchuett:
Implementing the subset function equivalents for RDMs as for dataset


iancharest (changes requested):
Hi Heiko,
1)
in `def get_vectors(self):` 
the docstring is a bit confusing: 
```python
    """ Returns RDMs as np.ndarray with each RDM as a vector
        Returns:
            RDMs as with one vector as one RDM
    """
```
same for get_matrices
```python
        """ Returns RDMs as np.ndarray with each RDM as a matrix
        Returns:
            RDMs as with one matrix as one RDM
        """
```
"RDMs as with one matrix as one RDM" what does this mean? 

otherwise this looks good.

HeikoSchuett:
Hi, I now merged the master back in and fixed the help messages. Thank you for having a look!

iancharest (approved):
great!

eberlot (approved):

PR #48, opened 2020-02-28
RDM batch processing

HeikoSchuett:
this pull request shall implement calculating an RDMs object from a list of datasets. As usually this is not finished immediately


HeikoSchuett:
This was not that much code and refers to a lot of code I wrote elsewhere, will pull it with other pull requests

PR #49, opened 2020-03-05
fix pairwise contrast speed issue

HeikoSchuett:
with numpys sum function instead of generic sum leads to an immense speedup


iancharest (approved):
Good catch!

eberlot:
that's great!

PR #58, opened 2020-03-19
Fix crossnobis

HeikoSchuett:
This is a fix to pass through the cv_descriptor necessary for calc_rdm_crossnobis.
It also sets a basic default now

### DIFF
--- a/pyrsa/data/noise.py
+++ b/pyrsa/data/noise.py
@@ -39,15 +39,20 @@ def cov_from_residuals(residuals, dof=None):
     else:
         if dof is None:
             dof = residuals.shape[0] - 1
+        # calculate sample covariance matrix s
         residuals = residuals - np.mean(residuals, axis=0, keepdims=True)
         xt_x = np.einsum('ij, ik-> ijk', residuals, residuals)
         s = np.sum(xt_x, axis=0) / xt_x.shape[0]
+        # calculate the scalar estimators to find the optimal shrinkage:
+        # m, d^2, b^2 as in Ledoit & Wolfe paper
         m = np.sum(np.diag(s)) / s.shape[0]
         d2 = np.sum((s - m * np.eye(s.shape[0])) ** 2)
         b2 = np.sum((xt_x - s) ** 2) / xt_x.shape[0] / xt_x.shape[0]
         b2 = min(d2, b2)
+        # shrink covariance matrix
         s_shrink = b2 / d2 * m * np.eye(s.shape[0]) \
             + (d2-b2) / d2 * s
+        # correction for degrees of freedom
         s_shrink = s_shrink * xt_x.shape[0] / dof
     return s_shrink
 
--- a/pyrsa/data/noise.py
+++ b/pyrsa/data/noise.py
@@ -10,7 +10,7 @@ from collections.abc import Iterable
 import numpy as np
 
 
-def get_cov_from_residuals(residuals, df=None):
+def get_cov_from_residuals(residuals, dof=None):
     """
     computes an optimal shrinkage estimate of the precision matrix from 
     the residuals as described by Ledoit and Wolfe (2004): "A well-conditioned
@@ -19,7 +19,7 @@ def get_cov_from_residuals(residuals, df=None):
     Args:
         residuals(numpy.ndarray or list of these): n_obs x n_channels matrix
             of residuals
-        df(int or list of int): degrees of freedom for covariance estimation
+        dof(int or list of int): degrees of freedom for covariance estimation
             defaults to n_obs-1, should be corrected for 
 
     Returns:
@@ -29,15 +29,15 @@ def get_cov_from_residuals(residuals, df=None):
     if not isinstance(residuals, np.ndarray) or len(residuals.shape) > 2:
         s_shrink = []
         for i in range(len(residuals)):
-            if df is None:
+            if dof is None:
                 s_shrink.append(get_cov_from_residuals(residuals[i]))
-            elif isinstance(df, Iterable):
-                s_shrink.append(get_cov_from_residuals(residuals[i], df[i]))
+            elif isinstance(dof, Iterable):
+                s_shrink.append(get_cov_from_residuals(residuals[i], dof[i]))
             else:
-                s_shrink.append(get_cov_from_residuals(residuals[i], df))
+                s_shrink.append(get_cov_from_residuals(residuals[i], dof))
     else:  
-        if df is None:
-            df = residuals.shape[0] - 1 
+        if dof is None:
+            dof = residuals.shape[0] - 1 
         residuals = residuals - np.mean(residuals, axis=0, keepdims=True)
         xt_x = np.einsum('ij, ik-> ijk', residuals, residuals)
         s = np.sum(xt_x, axis=0) / xt_x.shape[0]
@@ -47,11 +47,11 @@ def get_cov_from_residuals(residuals, df=None):
         b2 = min(d2, b2)
         s_shrink = b2 / d2 * m * np.eye(s.shape[0]) \
             + (d2-b2) / d2 * s
-        s_shrink = s_shrink * xt_x.shape[0] / df
+        s_shrink = s_shrink * xt_x.shape[0] / dof
     return s_shrink
 
 
-def get_prec_from_residuals(residuals, df=None):
+def get_prec_from_residuals(residuals, dof=None):
     """
     computes an optimal shrinkage estimate of the precision matrix from 
     the residuals as described by Ledoit and Wolfe (2004): "A well-conditioned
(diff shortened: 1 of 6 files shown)


pyrsa/data/noise.py
```diff
@@ -0,0 +1,76 @@
...
+Created on Mon Apr 13 12:42:06 2020
+
+@author: heiko
+"""
+
+from collections.abc import Iterable
+import numpy as np
+
+
+def get_cov_from_residuals(residuals, df=None):
```
JasperVanDenBosch: instead of `get_` I'd maybe just name them `cov_from` etc. This removes a step e.g. when on the commandline and looking for functions using <TAB>. But in any case the same naming strategy should be applied everywhere.

pyrsa/data/noise.py
```diff
@@ -0,0 +1,76 @@
...
+        residuals(numpy.ndarray or list of these): n_obs x n_channels matrix
+            of residuals
+        df(int or list of int): degrees of freedom for covariance estimation
+            defaults to n_obs-1, should be corrected for 
+
+    Returns:
+        numpy.ndarray (or list): sigma_p: covariance matrix over channels
+
+    """
+    if not isinstance(residuals, np.ndarray) or len(residuals.shape) > 2:
```
JasperVanDenBosch, line 30: Isn't it better to return a list either way, with just one element?

pyrsa/data/noise.py
```diff
@@ -0,0 +1,76 @@
...
+def get_prec_from_residuals(residuals, df=None):
+    """
+    computes an optimal shrinkage estimate of the precision matrix from 
+    the residuals as described by Ledoit and Wolfe (2004): "A well-conditioned
+    estimator for large-dimensional covariance matrices"
+
+    Args:
+        residuals(numpy.ndarray or list of these): n_obs x n_channels matrix
+            of residuals
+        df(int or list of int): degrees of freedom for covariance estimation
```
JasperVanDenBosch: `df` is also often used as an abbreviation of `DataFrame`

pyrsa/rdm/calc.py
```diff
@@ -130,24 +141,24 @@ def calc_rdm_mahalanobis(dataset, descriptor=None, noise=None):
...
-               descriptors=dataset.descriptors)
-    rdm.pattern_descriptors[descriptor] = desc
-    rdm.descriptors['noise'] = noise
+    if noise is None:
+        rdm = calc_rdm_euclid(dataset, descriptor)
+    else:
+        measurements, desc, descriptor = _parse_input(dataset, descriptor)
+        noise = _check_noise(noise, dataset.n_channel)
+        diff = _calc_pairwise_differences(measurements)
+        diff2 = (noise @ diff.T).T
```
JasperVanDenBosch, line 155: perhaps a comment here about what `diff` and `diff2` are

pyrsa/rdm/calc.py
```diff
@@ -130,24 +141,24 @@ def calc_rdm_mahalanobis(dataset, descriptor=None, noise=None):
...
 
 
-def calc_rdm_crossnobis(dataset,
-                        descriptor,
-                        noise=None,
-                        cv_descriptor=None
-                        ):
+def calc_rdm_crossnobis(dataset, descriptor, noise=None,
+                        cv_descriptor=None, subversion='whiten'):
     """
```
JasperVanDenBosch, line 167: I think it rightly warns here about the CC of 34 I think. It looks like this is currently tested with this:

```
    def test_calc_crossnobis(self):
        rdm = rsr.calc_rdm_crossnobis(self.test_data, descriptor = 'conds', cv_descriptor = 'fold')
        assert rdm.n_cond == 6
```

So this probably only tests one path through the function. Just think of all the emails you're gonna get "I did this and then I got this error, what's wrong?"

pyrsa/util/data_utils.py
```diff
@@ -3,10 +3,8 @@
 """
 Collection of helper methods for data module
 
-+ check_dict_length:   check if each value in dict matches length n
 + extract_dict:        extract key-value pairs with values given indexes.
 + get_unique_unsorted: return a unique unsorted list
-+ check_descriptors_dimension: run check_dict_length and raise an error
 
 @author: baihan
```
JasperVanDenBosch, line 9: So you should add yourself to the author list here if we are still doing that. If not let's get rid of them

JasperVanDenBosch (changes requested):

pyrsa/data/noise.py
HeikoSchuett: That is true, renamed to dof, which should not have this confusion

pyrsa/data/noise.py
HeikoSchuett: is renamed

pyrsa/data/noise.py
HeikoSchuett: This variant returns a list with one element if you pass it a list with one element and returns an array if you pass an array. That can create annoyances if you mix things, but it seems consistent to me?

pyrsa/util/data_utils.py
HeikoSchuett: Well, I did not update them anywhere. I would vote for getting rid of them, but didn't do so because I thought some people liked to have them in. Let's discuss this elsewhere

pyrsa/rdm/calc.py
HeikoSchuett: That is a good point. I will write a few more tests now to get at least a bit more covered

HeikoSchuett: now I added some more tests. I will now click this as resolved, but if you have ideas how to make the crossnobis easier I think this would be interesting still

JasperVanDenBosch (approved):
thanks for looking into my suggestions

pyrsa/data/noise.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def cov_from_residuals(residuals, dof=None):
+    """
+    computes an optimal shrinkage estimate of the precision matrix from 
+    the residuals as described by Ledoit and Wolfe (2004): "A well-conditioned
+    estimator for large-dimensional covariance matrices"
+
+    Args:
+        residuals(numpy.ndarray or list of these): n_obs x n_channels matrix
```
adkipnis: please rename n_obs, such that it does not get confused with n_obs in dataset/rdm objects

pyrsa/rdm/calc.py
```diff
@@ -48,7 +58,8 @@ def calc_rdm(dataset, method='euclidean', descriptor=None, noise=None):
         elif method == 'mahalanobis':
             rdm = calc_rdm_mahalanobis(dataset, descriptor, noise)
         elif method == 'crossnobis':
-            rdm = calc_rdm_crossnobis(dataset, descriptor, noise)
+            rdm = calc_rdm_crossnobis(dataset, descriptor, noise,
+                                      cv_descriptor=cv_descriptor)
```
adkipnis: `calc_rdm_crossnobis(dataset, descriptor, noise, cv_descriptor) `
would be sufficient

pyrsa/data/noise.py
```diff
@@ -0,0 +1,78 @@
...
+            if dof is None:
+                s_shrink.append(cov_from_residuals(residuals[i]))
+            elif isinstance(dof, Iterable):
+                s_shrink.append(cov_from_residuals(residuals[i], dof[i]))
+            else:
+                s_shrink.append(cov_from_residuals(residuals[i], dof))
+    else:
+        if dof is None:
+            dof = residuals.shape[0] - 1
+        residuals = residuals - np.mean(residuals, axis=0, keepdims=True)
```
adkipnis, line 42: Add to docstring that residual data get de-meaned

adkipnis, line 43: you could describe the steps of the algebraic shrinkage transform to keep consistency (also a sanity check that you followed Ledoit and Wolf's method correctly)

pyrsa/rdm/calc.py
```diff
@@ -130,27 +141,43 @@ def calc_rdm_mahalanobis(dataset, descriptor=None, noise=None):
...
-    diff = _calc_pairwise_differences(measurements)
-    diff2 = (noise @ diff.T).T
-    rdm = np.einsum('ij,ij->i', diff, diff2) / measurements.shape[1]
-    rdm = RDMs(dissimilarities=np.array([rdm]),
-               dissimilarity_measure='Mahalanobis',
-               descriptors=dataset.descriptors)
-    rdm.pattern_descriptors[descriptor] = desc
-    rdm.descriptors['noise'] = noise
+    if noise is None:
+        rdm = calc_rdm_euclid(dataset, descriptor)
```
adkipnis, line 147: Add warning that a precision matrix needs to be provided and that you automatically revert to Euclidian distance estimates otherwise

adkipnis (changes requested):

pyrsa/rdm/calc.py
HeikoSchuett: I put a comment on that into the help strings. I don't want to raise a warning as this can be completely intentional, for example to calculate a cross-validated Euclidean distance.

pyrsa/data/noise.py
HeikoSchuett: done

adkipnis (approved):

PR #59, opened 2020-03-23
Save load

### LINKED ISSUE #57: Implement save and load functions
For datasets, RDMs and results

HeikoSchuett:
implements save and load functions for the main objects we create

closes #57

### DIFF
--- a/pyrsa/data/dataset.py
+++ b/pyrsa/data/dataset.py
@@ -153,6 +153,21 @@ class DatasetBase:
                      self.obs_descriptors,
                      self.channel_descriptors],
                     open(filename + '_desc.pkl','wb'))
+    def to_dict(self):
+        """ Generates a dictionary which contains the information to
+        recreate the dataset object. Used for saving to disc
+        
+        Returns:
+            data_dict(dict): dictionary with dataset information
+
+        """
+        data_dict = {}
+        data_dict['measurements'] = self.measurements
+        data_dict['descriptors'] = self.descriptors
+        data_dict['obs_descriptors'] = self.obs_descriptors
+        data_dict['channel_descriptors'] = self.channel_descriptors
+        data_dict['type'] =  type(self).__name__
+        return data_dict
 
 
 class Dataset(DatasetBase):
@@ -276,3 +291,31 @@ def load_dataset(filename):
                    obs_descriptors=desc[1],
                    channel_descriptors=desc[2])
     return data
+
+
+def dataset_from_dict(data_dict):
+    """ regenerates a Dataset object from the dictionary representation
+
+    Currently this function works for Dataset and DatasetBase objects
+
+    Args:
+        data_dict(dict): the dictionary representation
+    
+    Returns:
+        data(Dataset): the regenerated Dataset
+
+    """
+    if data_dict['type'] == 'Dataset':
+        data = Dataset(data_dict['measurements'],
+                       descriptors=data_dict['descriptors'],
+                       obs_descriptors=data_dict['obs_descriptors'],
+                       channel_descriptors=data_dict['channel_descriptors'])
+    elif data_dict['type'] == 'DatasetBase':
+        data = DatasetBase(data_dict['measurements'],
+                       descriptors=data_dict['descriptors'],
+                       obs_descriptors=data_dict['obs_descriptors'],
+                       channel_descriptors=data_dict['channel_descriptors'])
+    else:
+        raise ValueError('type of Dataset not recognized')
+    return data
+        
--- a/pyrsa/data/dataset.py
+++ b/pyrsa/data/dataset.py
@@ -14,6 +14,10 @@ from pyrsa.util.descriptor_utils import subset_descriptor
 from pyrsa.util.descriptor_utils import bool_index
 from pyrsa.util.descriptor_utils import format_descriptor
 from pyrsa.util.descriptor_utils import parse_input_descriptor
+from pyrsa.util.file_io import write_dict_hdf5
+from pyrsa.util.file_io import write_dict_pkl
+from pyrsa.util.file_io import read_dict_hdf5
+from pyrsa.util.file_io import read_dict_pkl
 
 
 class DatasetBase:
@@ -139,20 +143,23 @@ class DatasetBase:
         """
         raise NotImplementedError(
             "subset_channel function not implemented in used Dataset class!")
-    def save(self, filename):
-        """ Saves the dataset object to two files:
-            [filename].npy : the data in numpy format
-            [filename]_desc.pkl : pickle of the descriptors
+    def save(self, filename, file_type='hdf5'):
+        """ Saves the dataset object to a file
             
         Args:
-            filename(String): path to the file(s)
+            filename(String): path to the file
+                [or opened file]
+            file_type(String): Type of file to create:
+                hdf5: hdf5 file
+                pkl: pickle file 
 
         """
-        np.save(filename + '.npy', self.measurements)
-        pickle.dump([self.descriptors,
-                     self.obs_descriptors,
-                     self.channel_descriptors],
-                    open(filename + '_desc.pkl','wb'))
+        data_dict = self.to_dict()
+        if file_type == 'hdf5':
+            write_dict_hdf5(filename, data_dict)
+        elif file_type == 'pkl':
+            write_dict_pkl(filename, data_dict)
+
     def to_dict(self):
         """ Generates a dictionary which contains the information to
         recreate the dataset object. Used for saving to disc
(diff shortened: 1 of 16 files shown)


HeikoSchuett:
As this contains the Results object this is based on the model-eval-plot branch, should merge after that one

.codeclimate.yml
```diff
@@ -2,16 +2,16 @@ version: "2"         # required to adjust maintainability checks
 checks:
   argument-count:
     config:
-      threshold: 7
+      threshold: 10
```
JasperVanDenBosch: You can also do `disabled: true` or something like that

pyrsa/data/dataset.py
```diff
@@ -137,6 +139,20 @@ def subset_channel(self, by, value):
         """
         raise NotImplementedError(
             "subset_channel function not implemented in used Dataset class!")
+    def save(self, filename):
+        """ Saves the dataset object to two files:
+            [filename].npy : the data in numpy format
+            [filename]_desc.pkl : pickle of the descriptors
```
JasperVanDenBosch: Have you considered using hdf5? Then you could store the descriptors as a dictionary or whatever in the same file. (So the user does not get confused about the two files). Also hdf5 is a bit more cross-platform (e.g. matlab).

pyrsa/inference/crossvalsets.py
```diff
@@ -10,34 +10,84 @@
 from pyrsa.util.rdm_utils import add_pattern_index
 
 
-def sets_leave_one_out_pattern(rdms, pattern_descriptor=None):
+def sets_leave_one_out_pattern(rdms, pattern_descriptor):
```
JasperVanDenBosch: How is this related to the save/load feature?

pyrsa/rdm/rdms.py
```diff
@@ -240,3 +252,79 @@ def subsample(self, by, value):
...
+        This function generates two files:
+            [filename].npy : a numpy array file to save the dissimilarities
+            [filename]_desc.pkl : a pickled file of the descriptors
+
+        Args:
+            filename(String): path to file to save to
+
+        """
+        np.save(filename + '.npy', self.dissimilarities)
+        pickle.dump([self.descriptors,
```
JasperVanDenBosch: If you go down this route, instead of a list, you could use a dictionary. This way you don't have to maintain the same order in the future and you'll save yourself headaches

tests/test_data.py
```diff
@@ -134,5 +134,30 @@ def test_average_by(self):
...
+                           obs_descriptors=obs_des,
+                           channel_descriptors=chn_des
+                           )
+        try:
+            data.save('test')
+            data_loaded = rsd.load_dataset('test')
+        finally:
+            import os
+            os.remove('test.npy')
+            os.remove('test_desc.pkl')
```
JasperVanDenBosch: I predict that this is gonna cause trouble.. It's good that it's in a finally loop, but that won't help against `ctrl-c`. Another way is to store them in an in-memory file object like `io.BytesIO`. If the filesystem interaction bit is a separate class you could have a MockFilesystem or something, or alternatively you could `@patch('my_module.open')`

JasperVanDenBosch: Maybe don't worry about this right now but keep it in mind if there's gonna be more file I/O

.codeclimate.yml
HeikoSchuett: So far this was enough. I actually agree with code climate to some degree that there should not be too many arguments, but with the inference stuff many functions need quite a bit of comments, so I felt this was a good idea

pyrsa/data/dataset.py
HeikoSchuett: Thanks for this comment. I am quite unsure what the best format is for saving here. I'll try out hdf5 this afternoon. If it can save all the content this would be better I think

pyrsa/rdm/rdms.py
HeikoSchuett: I like the dictionary idea. Will implement this soon

tests/test_data.py
HeikoSchuett: I hope that this pull request does most of the file i/o. The in memory idea sounds good to me. That would not easily with multiple files I guess or at least not as an actual test. I will come back to this once I tried the hdf5 idea. If this allows me to save stuff into one file, I should be able to make this test work in memory only. Thanks!

pyrsa/data/dataset.py
HeikoSchuett: hmmm hdf5 seems to imply limitations to the descriptors as we can only save arrays as datasets and strings as attributes. Dictionaries cannot be written directly to hdf5 apparently. I guess I could usually convert them such that they form groups which only contain strings and arrays, such that I can save them, but that honestly sounds like a bit of work to me and will not be entirely general, i.e. you could generate valid objects you cannot save.

I am somewhat tempted to just convert the object to a dictionary of dictionaries and np.arrays and pickle that as a whole. Do you know how wasteful this is compared to saving the main part of the object (i.e. the RDMs/measurements/evaluations) as hdf5 data and the rest as attribute annotations? Or is there any other reason to prefer hdf5 over the pickle, if I save only NumPy arrays and strings within layered dictionaries? For example: How bad is the security issue pickle mentions on their help?

HeikoSchuett: never mind. I have now started the conversions to dictionaries. I'll then write a function which writes dictionaries into pickles and one which writes them into hdf5. That should expose the choice to the user and if they used something which cannot be saved to hdf5 this will raise an error. That seems fair to me. Any objections?

HeikoSchuett: now implemented as described

tests/test_data.py
HeikoSchuett: now switched also to saving into `io.BytesIO` objects

tests/test_inference.py
```diff
@@ -218,13 +218,67 @@ def test_bootstrap_testset_rdm(self):
...
+        model_loaded = model_from_dict(model_dict)
+        assert m.name == model_loaded.name
+        assert np.all(m.rdm_obj.dissimilarities == model_loaded.rdm_obj.dissimilarities)
+
+        from pyrsa.model import ModelInterpolate
+        m = ModelInterpolate('test1', np.random.rand(10))
+        model_dict = m.to_dict()
+        model_loaded = model_from_dict(model_dict)
+        assert m.name == model_loaded.name
+        assert np.all(m.rdm_obj.dissimilarities == model_loaded.rdm_obj.dissimilarities)
```
JasperVanDenBosch, line 236: There's also `numpy.testing.assert_allclose` etc. These will give you failure messages that give a hint to what's going on

tests/test_inference.py
```diff
@@ -218,13 +218,67 @@ def test_bootstrap_testset_rdm(self):
         evaluations, n_rdms = bootstrap_testset_rdm([m, m2], rdms,
             method='cosine', fitter=None, N=100, rdm_descriptor=None)
 
-class test_save_load_result(unittest.TestCase):
+class test_save_load(unittest.TestCase):
```
JasperVanDenBosch: Class names by convention are written in *CamelCase*

pyrsa/util/file_io.py
```diff
@@ -0,0 +1,111 @@
...
+    Args:
+        file: a filename or opened writable file
+        dictionary(dict): the dict to be saved 
+
+    """
+    file = h5py.File(file, 'a')
+    _write_to_group(file, dictionary)
+    
+    
+def _write_to_group(group, dictionary):
```
JasperVanDenBosch, line 28: If the dicts only contain numbers and strings (settings) another option ofcourse is to serialize them with `ast`, or as tuples, or to json. I think this is was nibabel does with parameters in the Nifti header.

pyrsa/inference/result.py
```diff
@@ -47,36 +51,82 @@ def __init__(self, models, evaluations, method, cv_method, noise_ceiling):
...
 
         """
-        if isinstance(filename, str) and os.path.isfile(filename):
-            raise FileExistsError('File already exists')
-        file = h5py.File(filename, 'a')
-        file['evaluations'] = self.evaluations
-        file['noise_ceiling'] = self.noise_ceiling
-        file.attrs['method'] = self.method
-        file.attrs['cv_method'] = self.cv_method
-        file.attrs['version'] = 3
```
JasperVanDenBosch: We don't want to store the software version anymore?

pyrsa/data/dataset.py
```diff
@@ -14,6 +14,10 @@
 from pyrsa.util.descriptor_utils import bool_index
 from pyrsa.util.descriptor_utils import format_descriptor
 from pyrsa.util.descriptor_utils import parse_input_descriptor
+from pyrsa.util.file_io import write_dict_hdf5
```
JasperVanDenBosch, line 17: One way to make this a bit less verbose could be `import pyrsa.util.file_io as fio` and then `fio.write_dict_hdf5()`

JasperVanDenBosch (approved):
I'm happy that hdf5 is in there, and that we avoid writing files during testing. I've added some notes for future consideration but looks good to go to me. Thanks Heiko!

tests/test_inference.py
HeikoSchuett: Thanks for the reminder!

pyrsa/util/file_io.py
HeikoSchuett: Yes, I think I will leave this for now. The way I coded it this should be implementable as more functions in this file in the future

pyrsa/inference/result.py
HeikoSchuett: I actually include it again

eberlot (approved):
looks good!

PR #61, opened 2020-04-02
excluded nan-values from comparing rdms

### LINKED ISSUE #60: Bootstrapping over patterns-> nans instead of 0s
Previous toolboxes set values for comparing stimuli to themselves to nan/ignore instead of 0s as pyrsa does currently. This appears sensible to avoid inflated model performance, as all models will predict 0s for those comparisons.

HeikoSchuett:
implements the change from #60 
For subsampling over pattens now NaNs are produced for off-diagonal comparisons of the same patterns. The compare functions are now told to ignore these nan values.


eberlot (approved):
looks good!

nkriegeskorte (approved):

adkipnis (approved):

PR #62, opened 2020-04-02
Cov weighting

### LINKED ISSUE #53: implement Jörns new methods
I.e. at least the correlations which take the distribution for 0 rdms into account

HeikoSchuett:
This implements Jörn's covariance weighted comparisons between RDMs.

If you guys now faster ways of doing this, I would like to know ;)

### DIFF
--- a/pyrsa/rdm/calc.py
+++ b/pyrsa/rdm/calc.py
@@ -9,7 +9,7 @@ import numpy as np
 from pyrsa.rdm.rdms import RDMs
 from pyrsa.data.dataset import Dataset
 from pyrsa.data import average_dataset_by
-from pyrsa.util.matrix import pairwise_contrast
+from pyrsa.util.matrix import pairwise_contrast_sparse
 
 
 def calc_rdm(dataset, method='euclidean', descriptor=None, noise=None):
@@ -63,8 +63,8 @@ def calc_rdm_euclid(dataset, descriptor=None):
 
     """
     measurements, desc, descriptor = _parse_input(dataset, descriptor)
-    c_matrix = pairwise_contrast(np.arange(measurements.shape[0]))
-    diff = np.matmul(c_matrix, measurements)
+    c_matrix = pairwise_contrast_sparse(np.arange(measurements.shape[0]))
+    diff = c_matrix @ measurements
     rdm = np.einsum('ij,ij->i', diff, diff) / measurements.shape[1]
     rdm = RDMs(dissimilarities=np.array([rdm]),
                dissimilarity_measure='euclidean',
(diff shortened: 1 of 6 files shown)


jdiedrichsen:
Looks good overall. 
Couple of quick questions, though: 
a) sigma_k needs to be given as a matrix, instead of a 2D np-array. Would it not be more consistent to stick with 2D np-arrays throughout? If not, what's our rule to determine matrix vs. np array. 
b) Just curious; You use einsum for relatively simple inner products - is this easier or faster - as for sure it's not more readable! If we keep it, we probably should write the mathematical operations out explicitly. 
c) The weighted correlation is implemented by subtracting out the baseline before prewhitening- I guess that this is correct, but I'm just checking, as we haven't made this 100% clear in the paper
d) For flexible model, we will need a fast V^{-1/2} calculation. We can leave this for the next commit, but it's worth thinking about

jdiedrichsen (approved):
Otherwise, looks good

HeikoSchuett:
Hi Jörn!
Thanks for your comments!
a) The need to pass a matrix resulted from me using sparse matrices internally. That was not intended to be visible to the user. I now just put in a conversion of your matrix such that the function now accepts a numpy array for sigma_k
b) That is indeed a speed/notation issue. The Einstein sum allows us to compute the inner products for sets of vectors in one go. This would require loops and reshapes etc. to write out as an iteration. Since I got used to the notation I actually find them quite readable.
Ian checked this for the original cosine & correlation distances and saw a speed improvement as well. 
I will for now add some comments to make this clearer
c) I think this should be right. As the prewhitening is linear subtracting the mean before or after it is the same, right? I chose this path here for coding convenience reusing the cosine distance.
d) Yes, I would definitely like to talk to you about this and efficiency in general.
As I implemented it now this is ok for computing this for the relatively large matrices which occur for example for the 92 images, but for optimizing it could be worth computing the inverse of V or a V^(-1/2). Also given the structure of the matrix there might be some better ways of computing stuff. 
I am not entirely happy with the speed yet.

jdiedrichsen:
Hi Heiko - 
I implemented the covariance-weighting with a direct transform into a vector of elements of the second moment matrix. To avoid having to call squareform for each RDM, I implemented the double centering over some new contrast matrices (row_col_indicator_G). I anticipate these will be very handy for other uses - and I think I can manage to extend this to the case of a specific Sigma_K. Speed improvement is about 9x over the other option - relatively stable across matrix sizes. 
I would strongly expect that moving to sparse indicator matrices will bring additional speed improvements. 

cosine_cov_weighted_fast is not integrated in the code. I left it for you to see before integrating it in the main code.

eberlot:
this looks good to me!

HeikoSchuett:
Hi @eberlot , could you confirm this with a review you reach under "Files changed"->"Review changes"?

eberlot (approved):
Looks good to me!

PR #65, opened 2020-04-15
Rho_a aka spearman_rae implementation

### LINKED ISSUE #64: implement spearman_a / random among equals
The new/old correlation coefficient for rank correlations

HeikoSchuett:
This implements the rho_a or spearman random among equals comparison for a rank correlation which does not prefer models with tied ranks

### DIFF
--- a/pyrsa/data/__init__.py
+++ b/pyrsa/data/__init__.py
@@ -1,3 +1,5 @@
 from .dataset import Dataset
+from .dataset import load_dataset
+from .dataset import dataset_from_dict
 from .computations import average_dataset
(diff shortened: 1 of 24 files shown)


HeikoSchuett:
closes #64

nkriegeskorte (approved):

JasperVanDenBosch (approved):
I've not looked at the utilities changes, but the `rho_a` stuff looks good. I've made the unit test for "boot noise ceiling" parameterized. This means a separate test is run for each of the methods, meaning you can see which one failed right away.

PR #71, opened 2020-07-06
Updated documentation skeleton

jdiedrichsen:
I updated the structure of the Sphinx documentation in docs/source. 

The documentation now includes introductory pages and detailed documentation of each step in the RSA, which will ultimately result in a full manual for the toolbox. 

Examples of how to include figures, formulas, etc are included. 

I hope we can merge this quickly into the master, so that all developers can write the detailed documents and manual while programming.


nkriegeskorte:
That’s amazing.

JasperVanDenBosch:
And it's still compatible with `autogen` & `autodoc`? Where do you inform it of the new structure?

edit:

I've just tried to run the auto-generation (as in the readme file) on top of this but it reverts to the old structure. The API documentation follows the structure of the sub-packages themselves. I think it's a good idea to have some manual documentation as well for things like getting started guides, but I think it's more likely that we can get developers to write a quick docstring in the code, which is then turned into API documentation by `autogen`, then asking them to write RST for every new function they create. So I propose to keep the new narrated documentation but in addition to keep the autogenerated API documentation as well (`docs/source/pyrsa.rst` etc)

jdiedrichsen:
I am not using sphinx-autodoc and sphinx-autogen after the initial pass. Having the Module (which is empty) and Submodule stubs is tedious and looks awful. So I manually edited the rst files to fix sequence and structure to achieve a flatter and clearer hierarchy.  
The comments from the submodules are still being pulled automatically from the code through the statement in the respective .rst file, i.e.: 
.. automodule:: pyrsa.data.dataset
   :members:
   :undoc-members:
   :show-inheritance:

separately for each submodule. 

To build the documentation locally, I just call 

sphinx-build -a -b html source build 

This builds the html target (as it appears on readthedocs), but will not change the source files. 

Not sure how to tell readthedocs to not call sphinx-autogen, but there should be an option for it somewhere.

JasperVanDenBosch:
> Not sure how to tell readthedocs to not call sphinx-autogen, but there should be an option for it somewhere.

It can certainly be turned off, if that's what we want. Personally I'm opposed, but if there's another voter in favour for abandoning the autogenerated docs, then so be it. Apart from that I have no objections to this PR

HeikoSchuett:
First of all: Great that we got the documentation started!
I prefer the less deep tree for the documentation as well. It was always tedious to click into the individual files to find the functions. 
However, I much prefer the documentation to be autogenerated from the code. So I am only in favor if we can get readthedocs to parse this without rearranging the tree.

What are your opinions?

JasperVanDenBosch:
> if we can get readthedocs to parse this without rearranging the tree.

To clarify, it's not readthedocs that calls the apidoc extension, it's what the developers (are supposed to) do after
they add a module or otherwise change the structure of the package: https://github.com/rsagroup/pyrsa/blame/master/docs/README.md#L4

HeikoSchuett:
Ok, so we would have to produce the structure by hand after adding the documentation instead of running the automatic collection of things, but we would gain the option to change the tree structure any way we want. Do I understand that right? If so, I would be up for that.

As we are tiding this up: Should we drop the submodule layer? My understanding was that we use the __init__ files such that the file structure is essentially invisible to the user. It seems nicer to me to give the documentation the same structure, i.e. place the classes and functions directly below the modules.

JasperVanDenBosch (approved):

adkipnis (approved):

PR #74, opened 2020-07-07
data-overwrite

adkipnis:
added option to overwrite datafiles when saving (over data.save method)


pyrsa/util/file_io.py
```diff
@@ -117,3 +117,14 @@ def read_dict_pkl(file):
...
     return data
+
+def remove_file(file):
+    """ Deletes file from OS if it exists
+    
+    Args:
+        file: a filename or opened readable file
+
+    """
+    if isinstance(file, str) and os.path.exists(file):
```
HeikoSchuett: It would be great to make this slightly more general to work with non-string file identifiers or be explicit about them ignoring the overwrite option.
You would get something like this by using fopen for example.

adkipnis: Done!

HeikoSchuett (approved):

pyrsa/util/file_io.py
```diff
@@ -115,3 +115,17 @@ def read_dict_pkl(file):
...
     data = pickle.load(file)
     return data
+
+
+def remove_file(file):
+    """ Deletes file from OS if it exists
+
+    Args:
+        file (str, Path):
+            a filename or opened readable file
```
Benjamin-Peters, line 125: I am not sure this will work with an opened readable file as this is neither a str or Path instance. Specifically, remove_file(open('myrdm.pkl', 'rb')) would not work as far as I can see. I am not sure whether this is what is meant by "a filename or opened readable file".

Benjamin-Peters (changes requested):
I think it would be great to add tests for the remove_file function and the overwrite option of rdms and Dataset using both strings and opened files as filename.

In addition, it would generally be great to have tests for storing and loading data (if they do not exist already). This might be another pull request though.

pyrsa/util/file_io.py
HeikoSchuett: Hi Ben, I think it was meant to work on an opened file as well. I now added lines to make this happen.

PR #78, opened 2020-07-21
MultiArrangement task data input

### LINKED ISSUE #96: fix RDMs.sort_by()
The current version refers to a property called `measurements` which does not seem to be in use.

And when this is removed it is unclear how the following line:

```
dissimilarities[:, order][:, :, order]
```

would sort the pattern order

JasperVanDenBosch:
This aims to enable easy importing of behavioral results from the MA task as recorded on Meadows. For now it would support the *.mat* file type for this task, and all the scopes (single task / multiple tasks / one participant / multiple participants). [More on different file types here](https://meadows-research.com/documentation/researcher/downloads/). Future work could support other tasks with similarity data, such as the dragrate, triplet and discriminability task.

Here is an example of the interface:

```python
## general interface for all files, but denotes return type is RDMs object (to distinguish from DataSet TBA)
rdms = pyrsa.io.meadows.load_rdms('/path/to/ma_1D.mat')
```

#### Features / todo’s

- [x] detection of scope 
- [x] creating RDMs
- [x] scope: single task
- [x] scope: one participant / single task 
- [x] scope: multiple participants / single task
- [x] add descriptors from subject
- [x] sphinx auto doc stub
- [x] sample data files (neurdyn study? ~~Check for PIDs~~)
- [x] acceptance test: import sample file and check dataset similarity correspondence

#### revision

- [x] (optional flag) sort stimuli alphabetically and reorder RDM accordingly
- [x] resolves #96
- [x] Double check fortran/C ordering in MAT files, why is order z-a in test


HeikoSchuett (approved):
This seems to do what it is supposed to do and the tests confirm that it works, so I approve it.

I am not sure how general this is compared to the possible formats you could get from meadows, but I assume you do and it is general enough?

JasperVanDenBosch:
> I am not sure how general this is compared to the possible formats you could get from meadows, but I assume you do and it is general enough?

No to be honest this only supports one file type and only data for one type of task. But I figured it might be good to get some feedback on this now and at least have the most commonly used one in here, as the users often ask for code to do this.

ilogue:
- sort stimuli
- check order mat file

iancharest:
once we have the order of the stimuli (and rows and columns of the RDMs) sorted, so that this is then common across participants, i think i can approve this PR indeed.

JasperVanDenBosch:
@iancharest the rdm is now sorted by default, and I've compared the json and matlab order and they are the same.

JasperVanDenBosch:
@HeikoSchuett So I changed `RDMs.sort_by()`, what do you think?

iancharest (approved):
This now looks good. happy for this to be merged.

PR #86, opened 2020-08-06
Renaming

HeikoSchuett:
just renaming a few things to make it easier to read for users.

First a few things Niko noticed


doerlbh:
I am not quite sure what is the purpose of this PR. 

The author tags in each individual pyRSA files are not that troublesome, and on the contrary, can (1) let the users have a sense of the people behind the codes, such that copyright infringement is less likely, (2) let the users know who to ask specifically for any function, and (3) engage new contributors who added some auxiliary functionalities with some attributions and mental rewards (even if they might not be core contributors or acknowledged authors). 

Comparing to the author tags, the AUTHORS file in the outermost folder has a much more constrained effect (even 1-commit contributors goes into the "AUTHORS", and alphabetically... Are the "authors" the same as "contributors"? If so, I am also an *author* of scikit-learn, but am I really?)

The author tags are also a convention in popular Python packages. if you have a look at scikit-learn (sklearn) and numpy, at each individual modules, there are core developers/maintainers noted inside the python files. (That is also how I figure out how to properly attribute my modified MDS implementation to the original sklearn.mainifold.mds author.)

HeikoSchuett:
Hi Baihan,
so first removing these tags was not the main aim of this pull request at all. Purpose of changing the headers was to unify them. Different files had completely different formats. Here I remove the ones which just said that I created them at some timepoints and removed the listing of functions in the file to remove maintenance effort as these lists were not correct for most files I looked at anyway. 

My main aim for the authors tag is that I do not have to police them in any way though. Thus, if you want your tag there feel free to keep it. Give me a second to go through this pull request.

JasperVanDenBosch (approved):
Great to see uniform module headers!

doerlbh:
>  so first removing these tags was not the main aim of this pull request at all. Purpose of changing the headers was to unify them. Different files had completely different formats. Here I remove the ones which just said that I created them at some timepoints and removed the listing of functions in the file to remove maintenance effort as these lists were not correct for most files I looked at anyway.

Thanks for the clarifications, Heiko! I have a better understanding of the goal of this PR and am fully on board with the uniform module headers.

doerlbh (approved):

PR #87, opened 2020-08-06
Merge datasets after subsetting; perform odd-even split 

adkipnis:
Added function to merge lists of datasets with matching descriptors (as produced by subset_* methods). Added functions to perform odd-even splits on dataset objects according to up to two provided observation descriptors.


pyrsa/data/dataset.py
```diff
@@ -349,3 +349,126 @@ def dataset_from_dict(data_dict):
...
+    assert list(dict_orig.keys()) == list(dict_addit.keys()), "Provided observation descriptors have different keys."
+    
+    dict_merged = {}
+    keys = list(dict_orig.keys())
+    for k in keys:
+        values = np.array(np.append(dict_orig[k], dict_addit[k]))
+        dict_merged.update({k:values})
+    return dict_merged
+
+def odd_even_split(dataset, obs_desc, sort_by = None):
```
HeikoSchuett: I think this one can probably be done more efficiently by splitting the descriptor list and running two subset operations on the original dataset?

HeikoSchuett: Also: Do we need the sort_by option? I mean you could just run the sort command afterwards, right?

adkipnis:
I‘m not sure what you mean by running two subset operations on the original dataset. Could you please express this in pseudocode?

adkipnis:
I find it handy but I wouldn‘t mind removing the sort_by option.

pyrsa/data/dataset.py
```diff
@@ -349,3 +349,126 @@ def dataset_from_dict(data_dict):
...
+    odd_split = merge_subsets(odd_list)
+    even_split = merge_subsets(even_list)
+    
+    if sort_by is not None:
+        odd_split.sort_by(sort_by)
+        even_split.sort_by(sort_by)
+    return odd_split, even_split
+
+
+def nested_odd_even_split(dataset, l1_obs_desc, l2_obs_desc, sort_by = None):
```
HeikoSchuett: I am not sure what this is for. Is this for balancing or something like that? How is the resulting dataset different from running the single split half with l2_obs_desc?

HeikoSchuett:
So in general, I like the merging function and a split half might also be handy in many circumstances.

Please add some tests that this is working as intended!
And I think for code style shorter lines would be great.

pyrsa/data/dataset.py
adkipnis: Exactly, it‘s for balancing, especially if the order of your measurements is inconsistent, or if the two descriptors are not orthogonal.

pyrsa/data/dataset.py
adkipnis: I‘m not sure what you mean by running two subset operations on the original dataset. Could you please express this in pseudocode?

adkipnis: I find it handy but I wouldn‘t mind removing the sort_by option.

pyrsa/data/dataset.py
```diff
@@ -349,3 +349,126 @@ def dataset_from_dict(data_dict):
...
+        obs_descriptors = append_obs_descriptors(obs_descriptors, ds.obs_descriptors.copy())
+    
+    merged_dataset = Dataset(measurements,
+                                descriptors = descriptors, 
+                                obs_descriptors = obs_descriptors,                             
+                                channel_descriptors = channel_descriptors)
+    return merged_dataset
+
+
+def append_obs_descriptors(dict_orig, dict_addit):
```
HeikoSchuett: For consistency: Could you move this into the descriptor utils file?

pyrsa/data/dataset.py
HeikoSchuett: Sure, it would be something like this:
     even_values =  np.unique(dataset.obs_descriptors[obs_desc])[0::2]
     odd_values = np.unique(dataset.obs_descriptors[obs_desc])[1::2]
even_split = dataset.subset(obs_desc, even_values)
odd_split = dataset.subset(obs_desc, odd_values)

pyrsa/data/dataset.py
HeikoSchuett: Then maybe mention this in the docstring

pyrsa/data/dataset.py
```diff
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 """
 Definition of RSA Dataset class and subclasses
-@author: baihan, jdiedrichsen
+@author: baihan, jdiedrichsen, adkipnis
```
JasperVanDenBosch, line 6: Should we really go down this path?

https://stackoverflow.com/questions/17269843/javadoc-author-tag-good-practices

tests/test_data.py
```diff
@@ -206,6 +206,128 @@ def test_save_load(self):
...
+            measurements=measurements,
+            descriptors=des,
+            obs_descriptors=obs_des,
+            channel_descriptors=chn_des
+            )
+    
+    def test_merge(self):
+        subsets = self.test_data.split_obs('conds')
+        self.test_data_merged = rsd.merge_subsets(subsets)
+        assert np.all(self.test_data_merged.measurements == \
```
JasperVanDenBosch: please use `numpy.testing` assertions for these as they give very nice informative failure messages whereas this will just say `AssertionError`

tests/test_data.py
```diff
@@ -206,6 +206,128 @@ def test_save_load(self):
...
+        assert np.all(self.even_data.measurements == \
+                self.even_split.measurements)
+        assert self.even_data.descriptors == \
+                self.even_split.descriptors
+        assert np.all(self.even_data.obs_descriptors['conds'] == \
+                self.even_split.obs_descriptors['conds'])
+        assert np.all(self.even_data.channel_descriptors['rois'] == \
+                self.even_split.channel_descriptors['rois'])     
+
+class TestNestedOESplit(unittest.TestCase):
```
JasperVanDenBosch: I'd argue for one TestCase class per module

pyrsa/util/descriptor_utils.py
```diff
@@ -149,3 +149,19 @@ def check_descriptor_length_error(descriptor, name, n):
...
+    """
+    assert list(dict_orig.keys()) == list(dict_addit.keys()), \
+        "Provided observationdescriptors have different keys."
+    
+    dict_merged = {}
+    keys = list(dict_orig.keys())
+    for k in keys:
+        values = np.array(np.append(dict_orig[k], dict_addit[k]))
+        dict_merged.update({k:values})
+    return dict_merged
```
JasperVanDenBosch: missing newline at end of file here

pyrsa/data/dataset.py
```diff
@@ -349,3 +350,123 @@ def dataset_from_dict(data_dict):
...
+        odd_split (Dataset):
+            subset of the Dataset with odd list-indices after partitioning
+            according to obs_desc
+        even_split (Dataset):
+            subset of the Dataset with even list-indices after partitioning
+            according to obs_desc
+
+    """
+    assert l1_obs_desc and l2_obs_desc in dataset.obs_descriptors.keys(), \
+        "obs_desc must be contained in keys of dataset.obs_descriptors"
```
JasperVanDenBosch: Maybe write this out as "observation descriptors"?

pyrsa/data/__init__.py
```diff
@@ -1,6 +1,9 @@
 from .dataset import Dataset
 from .dataset import load_dataset
 from .dataset import dataset_from_dict
+from .dataset import merge_subsets
+from .dataset import odd_even_split
```
JasperVanDenBosch: Would it perhaps make sense to have these available as static methods on the `DataSet` class? That way the user won't have to import so many functions

JasperVanDenBosch (changes requested):
Looks mostly good, a few structural comments

pyrsa/data/dataset.py
adkipnis: Which ever path we choose, should be walked on consistently. I believe either all or none of the authors of a script should be mentioned. Wouldn't mind if we removed `@author` tags altogether.

pyrsa/data/__init__.py
adkipnis: As the merge_subsets function takes in lists of datasets, it's more useful as a function than as a method. I'll experiment with writing odd_even_split as a method.

adkipnis: Okay, this should work now.

adkipnis:
Thanks for the feedback, @ilogue !

pyrsa/data/dataset.py
HeikoSchuett: I agree with Jasper here. I think these author lines should be removed. Here on Github you can look up for any line who changed it last and what it was at any timepoint in the past. That should be enough.
As this is somewhat more related to the renaming stuff I included it there in #86 
For this pull request this should be irrelevant.

pyrsa/data/dataset.py
```diff
@@ -349,3 +421,123 @@ def dataset_from_dict(data_dict):
...
+#     odd_list = ds_part[0::2]
+#     even_list = ds_part[1::2]
+#     odd_split = merge_subsets(odd_list)
+#     even_split = merge_subsets(even_list)
+    
+    
+#     return odd_split, even_split
+
+
+# def nested_odd_even_split(dataset, l1_obs_desc, l2_obs_desc):
```
HeikoSchuett: Before merging into master, please remove all commented out code

JasperVanDenBosch (approved):
thanks for the changes

HeikoSchuett (approved):

PR #90, opened 2020-08-23
Changes and additions to calculating the precision matrix

adkipnis:
1. Added methods to datasets (getter method for measurements + method to create 3d array of measurements, where the data is split by an obs_descriptor along the third dimension).
2. Modularized precision matrix calculation
3. Added function to estimate covariance of a 3d array (averaging the covariance estimates along the third dimension)
4. Implemented the calculation of an instance-based covariance matrix, where the measurements are used instead of the residuals (as proposed by Tal Golan)
5. Minor changes in previous docstrings


pyrsa/data/dataset.py
```diff
@@ -299,7 +299,37 @@ def sort_by(self, by):
...
+    def get_measurements_tensor(self, by):
+        """ Returns a tensor version of the measurements array, split by an 
+        observation descriptor
 
+        Args:
+            by(String):
+                the descriptor by which the splitting is made
+
+        Returns:
+            measurements_tensor (ndarray):
```
HeikoSchuett: would leave a description which dimension is what

pyrsa/data/dataset.py
```diff
@@ -299,7 +299,37 @@ def sort_by(self, by):
...
+                3d array with the selected obs_descriptor constituting the
+                third dimension.
+        """
+        assert len(self.obs_descriptors.keys()) > 1, \
+            "dataset has too few obs_descriptors to build a tensor from it"
```
HeikoSchuett: Why couldn't I want to build a tensor from a dataset with only one descriptor?

pyrsa/data/dataset.py
```diff
@@ -299,7 +299,37 @@ def sort_by(self, by):
...
+            "third dimension not in obs_descriptors"
+        unique_values = get_unique_unsorted(self.obs_descriptors[by])
+        measurements_list = []
+        for v in unique_values:
+            selection = (self.obs_descriptors[by] == v)
+            measurements_subset = self.measurements[selection, :]
+            measurements_list.append(measurements_subset)
+        measurements_tensor = np.stack(measurements_list, axis=0)
+        measurements_tensor = np.swapaxes(measurements_tensor, 1, 2)
+        return measurements_tensor
```
HeikoSchuett: would it make sense to also return unique_values such that a user of this function could know which descriptor values correspond to which tensor slice?

pyrsa/data/dataset.py
```diff
@@ -299,7 +299,37 @@ def sort_by(self, by):
...
+                3d array with the selected obs_descriptor constituting the
+                third dimension.
+        """
+        assert len(self.obs_descriptors.keys()) > 1, \
+            "dataset has too few obs_descriptors to build a tensor from it"
+        assert by in self.obs_descriptors.keys(), \
+            "third dimension not in obs_descriptors"
+        unique_values = get_unique_unsorted(self.obs_descriptors[by])
+        measurements_list = []
+        for v in unique_values:
```
HeikoSchuett, line 330: I think this part expects that the data for each slice is in the same order. I would prefer a variant which ensures that the matching is not confused when we generate this.
One very thorough version of this would make the user specify all dimensions they want to have in their array and average over the rest.
As a minimal solution, tell the user that this just pastes measurements in the same order as they occur in the dataset.

HeikoSchuett (commented):
Overall this looks pretty good. I think it would be great if the get measurement tensor method could be more general & stable to against order issues, but that should be it

pyrsa/data/dataset.py
adkipnis: Because (unless you have multiple measurements with identical descriptors,) the dimensions of the tensor would be 1  x n_channels x n_obs_by.

HeikoSchuett: First: I think this line checks whether there multiple descriptors not whether there are multiple values. So, there are cases where I would end up with a perfectly fine array and cannot get it because this assert fails.
Second: Even if I end up with a 1 x n_channels x n_obs array that might well be what I want to do in some cases for example if the number of runs differs between subjects and some subjects have only one run.

HeikoSchuett (approved):
I just fixed the last few small issues, this should now be done

PR #91, opened 2020-08-31
Icon

HeikoSchuett:
Adds an Icon class which is an object, which once created can be plotted into an axis or placed as a tick marker.


doerlbh (approved):

doerlbh (commented):
Seems good me in a quick look. Would be good to check again when the unit tests and lint modules are reenabled https://github.com/rsagroup/pyrsa/issues/100.

JasperVanDenBosch (approved):

PR #95, opened 2020-09-10
Bugfix noise ceiling

HeikoSchuett:
This fixes a problem observed Baihan for the computation of the noise ceiling. 
The problem was that if no descriptor name was provided such a distinction was created newly for each bootstrap sample such that subjects could appear in both training and test set for the estimation of the lower bound on the noise ceiling, which inflated this bound. 

Please check and merge soon!

### DIFF
--- a/pyrsa/inference/noise_ceiling.py
+++ b/pyrsa/inference/noise_ceiling.py
@@ -11,6 +11,7 @@ from pyrsa.util.inference_util import pool_rdm
 from pyrsa.rdm import compare
 from .crossvalsets import sets_leave_one_out_rdm
 
+
 def cv_noise_ceiling(rdms, ceil_set, test_set, method='cosine',
                      pattern_descriptor=None):
     """ calculates the noise ceiling for crossvalidation.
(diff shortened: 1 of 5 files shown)


pyrsa/inference/noise_ceiling.py
```diff
@@ -42,10 +43,10 @@ def cv_noise_ceiling(rdms, ceil_set, test_set, method='cosine',
         train = ceil_set[i]
         test = test_set[i]
         pred_train = pool_rdm(train[0], method=method)
-        pred_train = pred_train.subsample_pattern(by=pattern_descriptor, 
+        pred_train = pred_train.subsample_pattern(by=pattern_descriptor,
                                                   value=test[1])
```
adkipnis, line 45: not completely sure but should this perhaps be `train[1] ` instead of ` test[1]` ?

adkipnis (approved):

pyrsa/inference/noise_ceiling.py
HeikoSchuett: I didn't change anything about this here, but test[1] are the patterns sampled for the test set. Thus in this line we choose from pred_train to the patterns in the test set. That seems sensible to me.

JasperVanDenBosch (approved):
It looks like the code became easier to follow which is always a good thing. In some cases it may be good to add a test which replicates the bug conditions, which helps to remember the requirement that it uncovered.

PR #102, opened 2020-09-17
fixing mds tests

HeikoSchuett:
the tests for the MDS functions fail
this shall fix this


doerlbh (approved):
Good point. Thanks!

doerlbh:
It makes sense that the results are not actually equal. As MDS embeddings can have different equivalent configurations (e.g., the dimensions don't matter if switched).

HeikoSchuett:
Then we need to check something else right?

doerlbh:
> Then we need to check something else right?

How do you think of comparing the pairwise distances computed from the MDS embeddings? e.g. assertEqual_allclose(pdist(mds_emb),pdist(wmds_emb))

doerlbh:
Their tests are all past on my end.

doerlbh:
Okay, now the unit tests are past.

HeikoSchuett:
Hi Baihan,
you now increased the tolerance quite a bit I think. The error can now be around .3. That seems substantial. Can you check what the difference from your implementation and the one you compare to is?

doerlbh:
It is the sum error (e.g. in 3D embedding of 5 condition, it is the sum of differences for 3x5x5=75 values, thus 0.3 absolute tolerance would be 0.3/75=0.004 for each value, which is very small). The sum absolute error that I got for the two pdist array is around 0.13 to 0.17 as you can run, thus I chose 0.3 as a reasonable threshold.

HeikoSchuett:
OK, can you check nonetheless whether one of the two is better in terms of matching the distances?

Otherwise I think we could merge this to avoid complains

I can't approve my own pull request...

JasperVanDenBosch (approved):
Maybe not the most straightforward way to test the MDS methods but this looks like a good fix

PR #112, opened 2020-09-17
Packaging for initial release

JasperVanDenBosch:
Simple configuration for a release on pypi


JasperVanDenBosch:
ready for review

HeikoSchuett:
Do you see any way to get the pyrsa name on pypi?
It seems to be an implementation of the RSA encryption algorithm which only exists there for a while so far. I cannot read much unfortunately because the text is Chinese...

JasperVanDenBosch:
> Do you see any way to get the pyrsa name on pypi?
> It seems to be an implementation of the RSA encryption algorithm which only exists there for a while so far. I cannot read much unfortunately because the text is Chinese...

Yeah I'm afraid the only way is to ask them if they want to give it up.. @doerlbh do you feel like asking them?

doerlbh:
> Yeah I'm afraid the only way is to ask them if they want to give it up.. @doerlbh do you feel like asking them?

PyPI's PyRSA is a RSA-based encryption package that was used in various education systems in China (where students, staffs and teachers can only view their individual accessible information under encryption). Regarding the name, although it would be great if we got the pyRSA name, I feel weird to ask them to give up their name. If we were in their shoes, how would you feel if someone else in a different field asking us to give up a software brand which has been released for almost a year? At least in my personal view, that would feel like a unreasonable request. What do you think?

JasperVanDenBosch:
> > Yeah I'm afraid the only way is to ask them if they want to give it up.. @doerlbh do you feel like asking them?
> 
> PyPI's PyRSA is a RSA-based encryption package that was used in various education systems in China (where students, staffs and teachers can only view their individual accessible information under encryption). Regarding the name, although it would be great if we got the pyRSA name, I feel weird to ask them to give up their name. If we were in their shoes, how would you feel if someone else in a different field asking us to give up a software brand which has been released for almost a year? At least in my personal view, that would feel like a unreasonable request. What do you think?

Yes I agree. I'm glad we got your input, I wasn't able to tell much from the repo. 

@HeikoSchuett is your main concern the incongruence with the code naming? Because in principle it's possible to rename everything to rsa3, or to keep the dual naming (as with scikit-learn / sklearn). Or you like the name `pyrsa` better?

HeikoSchuett:
No, indeed that sounds too big to get the name from them. I had hoped it was just some guys doing their own 10th version of RSA encryption in python. 
I am not a great fan of rsa3 as a name as rsa is the big encryption rsa toolbox right? That would definitely cause confusions.

I don't have a better solution though. so maybe we just take rsa3.

HeikoSchuett (approved):

doerlbh (approved):

PR #113, opened 2020-09-17
Temporal rsa

### LINKED ISSUE #24: Implement DatasetList options (including sliding window operations)
As discussed and seconded in pull request #16 , one design decision is to have a supplementary class called DatasetList object with the base of a list of dataset objects. This include implementation of other principled segmentation operations such as sliding window for dynamic analysis.

Benjamin-Peters:
Classes and functions to create RDM movies from temporal data


pyrsa/rdm/calc.py
```diff
@@ -464,6 +464,7 @@ def _gen_default_cv_descriptor(dataset, descriptor):
 
 
 def _calc_pairwise_differences(measurements):
+    measurements = measurements.squeeze()
```
Benjamin-Peters, line 466: If measurements come from DatasetTime, measurements is three-dimensional. The last dimension is empty here.
It may be worth checking whether this causes trouble if either n_obs or n_channels are one.

pyrsa/data/dataset.py
```diff
@@ -299,6 +299,286 @@ def sort_by(self, by):
         order = np.argsort(desc)
         self.measurements = self.measurements[order]
         self.obs_descriptors = subset_descriptor(self.obs_descriptors, order)
+        
+
+class DatasetTime(Dataset):
```
JasperVanDenBosch: So this is a minor, aesthetical thing but I tend to prefer ChildParent class naming,  e.g. DogMammal instead of MammalDog, so in this case perhaps `TimeDataset` or `TemporalDataset`? What do you think?

pyrsa/data/dataset.py
```diff
@@ -299,6 +299,286 @@ def sort_by(self, by):
...
+        obs_descriptors (dict):       observation descriptors (all
+            are array-like with shape = (n_obs,...))
+        channel_descriptors (dict):   channel descriptors (all are
+            array-like with shape = (n_channel,...))
+        time_descriptors (dict):      time descriptors (alls are 
+            array-like with shape= (n_time,...))
+        
+            time_descriptors needs to contain one key 'time' that 
+            specifies the time-coordinate. if None is provided, 'time' is
+            set as (0, 2, ..., n_time-1)
```
JasperVanDenBosch: Why is the second time point `2`?

JasperVanDenBosch (changes requested):
Made some suggestions, and two other things I'd like to see added are:

1. a method to convert it to a regular ("long-form") `Dataset`
2. I think you showed a demo notebook in the workshop

JasperVanDenBosch (approved):
I'm very happy with the changes and this PR is now ready to go afaic.

The issues codefactor is reporting are from other PRs. It ~~seems it's not smart enough to recognize master merges~~ picks up on issues in nearby code in the same file

HeikoSchuett (approved):
yes, this looks like a good starting point for RSA over time.

PR #118, opened 2020-09-18
RDMs implements __len__.

caiw:
We can already check the number of RDMs in a stack by using `RDMs.n_rdm`.  However, providing this through `__len__` has advantages:
- It is idiomatic: people will try to use `len(rdms)`, and this will give them what they're looking for.
- Together with `__getitem__`, which provides for iteration with `for rdm in rdms:`, `__len__` lets you use `reversed(rdms)`, which can be useful in some contexts.

### DIFF
--- a/pyrsa/rdm/rdms.py
+++ b/pyrsa/rdm/rdms.py
@@ -107,6 +107,7 @@ class RDMs:
     def __getitem__(self, idx):
         """
         allows indexing with []
+        and iterating over RDMs with `for rdm in rdms:`
         """
         idx = np.array(idx)
         dissimilarities = self.dissimilarities[idx].reshape(-1,
@@ -118,6 +119,13 @@ class RDMs:
                     rdm_descriptors=rdm_descriptors,
                     pattern_descriptors=self.pattern_descriptors)
         return rdms
+
+    def __len__(self) -> int:
+        """
+        The number of RDMs in this stack.
+        Together with __getitem__, allows `reversed(rdms)`.
+        """
+        return self.n_rdm
 
     def get_vectors(self):
         """ Returns RDMs as np.ndarray with each RDM as a vector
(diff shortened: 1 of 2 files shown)


HeikoSchuett (approved):
Looks good to me. I did not expect that this is already sufficient to do the iteration

caiw:
> Looks good to me. I did not expect that this is already sufficient to do the iteration

I didn't either! I just thoughtlessly typed `for rdm in rdms:` and was bewildered when it didn't raise an error… then I looked up why! :)

pyrsa/rdm/rdms.py
```diff
@@ -107,6 +107,7 @@ def __str__(self):
     def __getitem__(self, idx):
         """
         allows indexing with []
+        and iterating over RDMs with `for rdm in rdms:`
         """
         idx = np.array(idx)
```
JasperVanDenBosch, line 112: if the user tries to index with something that's not a valid array index, e.g. `rdms['a']`, it will raise:

```
IndexError: arrays used as indices must be of integer (or boolean) type
```

Should it also support indexing by descriptor? And is the error message clear enough?

JasperVanDenBosch (approved):
Great stuff! If I had to mention something it's that we may want to think on clarifying the `__getitem__` behavior a bit

pyrsa/rdm/rdms.py
caiw: This is a good question.  I also wondered about implementing slicing (e.g. `rdms[2:]`), `__setitem__` for things like `rdms[2:2] = RDMs(…)`, and `__delitem__`.  They seem like quite nice quality-of-life enhancements for objects which behave like lists of things.  But we'd need to be careful about when we return copies versus references, etc. I've made a new issue (#130) where we can discuss expanding indexing of RDMs.

PR #121, opened 2020-09-18
bug fix: avoid descriptor exceptions on 0d arrays (e.g. strings)

jooh:
check_descriptor_length casts each descriptor value to np.array. This can result in 0-d arrays if the input is str. The subsequent check on shape[0] then triggers an exception because 0-d arrays are weird and have no shape elements.

So I fixed that by flattening the array whenever the shape is empty. I also added a test.

Finally, I snuck in a stylistic change - single character variable names are rarely a good idea, and when the character clashes with a common pdb shortcut it's really ill advised.

### DIFF
--- a/pyrsa/util/descriptor_utils.py
+++ b/pyrsa/util/descriptor_utils.py
@@ -68,13 +68,13 @@ def parse_input_descriptor(descriptors):
     return descriptors
 
 
-def check_descriptor_length(descriptor, n):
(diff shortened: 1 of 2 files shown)


HeikoSchuett (approved):
thanks for writing a test immediately. looks good and I am happy to remove the n

JasperVanDenBosch (approved):
+1

PR #126, opened 2020-09-18
new code review settings

### LINKED ISSUE #33: Don't exclude tests from style checks
https://codeclimate.com/repos/5d9b9be58feef31572014b94/settings/exclude_patterns/edit

### LINKED ISSUE #100: Reenable unit test, codeclimate and lint checks for all PR commits
At this point, the automatic unit tests, codeclimate and lint style checks by GitHub doesn't seem to be working any more. Any experts looking into this would be appreciated. 

It would be great to have them as a coding guideline for contributors, especially because eventually we will officially launch this software package to the general public.

JasperVanDenBosch:
fixes #33 
resolves #100 

A number of configuration files to select which things we want to check for


JasperVanDenBosch:
So the tricky thing is that none of these apply until they're in master, so I think a second tuning step will be required after this is merged, and we can see what sort of issues they raise. The pylint settings are the default ones.

Ready for review

HeikoSchuett (approved):
Yes, I do not see a way to test this rc file for me, but adding these config files makes sense to me

doerlbh (approved):

PR #128, opened 2020-09-18
fixed descriptor issue in split_channel of Dataset class

Benjamin-Peters:
Without copy(), the descriptor[by] value of each dataset_list object is set to the last value of v in unique_values


HeikoSchuett (approved):
good catch

JasperVanDenBosch (approved):

PR #136, opened 2020-09-28
Align partial RDMs

JasperVanDenBosch:
In behavioral RSA it’s not uncommon to try to reconstruct a full RDM from a set of partial RDMs.
This is used for instance in the Multi Arrangement task to combine trials, but is also being considered now to combine individual subject RDMs into a “fixed-effects” population RDM, when the number of items is too large for each subject to judge each pair of items.
There are several existing matlab implementations for this, to be found in the MA task code base. For now I will aim to just implement the original one.

Proposed API:

1. First expand the measured rdms to have NaNs for missing pairs:
```python
partial_rdms = RDMs.expand([rdms1, rdms2, ..], descriptor=’cond’)
```

2. Then create the combined RDM:
```python
aligned_rdms = partial_rdms.align()
```

3. Then obtain the (optionally weighted) mean RDM:
```python
mean_rdm = aligned_rdms.mean()
```

4. Or as a chained expression:
```python
mean_rdm =  RDMs.expand([rdms1, rdms2, ..]).align().mean()
```

An alternative interface for 2) could be to store the new RDM in the existing RDMs object,
But looking at the existing functions, it seems the growing convention is to return a new object.

##### todos

- [x] test for `expand` from union
- [x] test for `expand` from chosen list
- [x] docstring for `expand`
- [x] implement `expand`
- [x] test for `align`
- [x] docstring for `align`
- [x] implement `align`
- [x] test for `mean`
- [x] docstring for `mean`
- [x] implement `mean`
- [x] narrative doc text for `expand` 

##### edits
- removed `count_missing()`. It would probably then make sense to have non-missing, and to do this by rdm as well as by pattern, at which point it may make more sense to expect the user to do this in numpy themselves
- split **mean** from **align** as the aligned rdms may be useful by themselves


lajnd (commented):
Looks good to me and clearly useful. Pylint is returning 10/10 :) Can't say much

lajnd (approved):

JasperVanDenBosch:


```diff

Complexity increasing per file
==============================
- pyrsa/rdm/align.py  4
- tests/test_rdms_align.py  1
- tests/test_rdms_expand.py  1
- pyrsa/rdm/rdms.py  8
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/rsagroup/pyrsa/pullRequest?prid=6247976&bid=20347491)

JasperVanDenBosch:
I'd really appreciate if someone could have a look at this..

pyrsa/rdm/rdms.py
```diff
@@ -374,6 +377,130 @@ def sort_by(self, **kwargs):
...
+            rdm_descriptors['weights'] = weights
+        return RDMs(
+            dissimilarities=aligned,
+            dissimilarity_measure=self.dissimilarity_measure,
+            descriptors=deepcopy(self.descriptors),
+            rdm_descriptors=rdm_descriptors,
+            pattern_descriptors=deepcopy(self.pattern_descriptors)
+        )
+
+    def mean(self, weights='stored'):
```
HeikoSchuett: I find the default of using stored weights a bit odd here.  I would certainly prefer the None version to be the default, i.e. the version which takes an unweighted average. rdms.mean() should really work. Also I am not sure, why self.descriptors['weights'] has to be removed. 

Would a version which accepts the name of the weights descriptor as input make sense? Such that the current behavior could be achieved by rdms.mean('weights') instead?

pyrsa/rdm/align.py
```diff
@@ -0,0 +1,83 @@
...
+        dissim (ndarray): dissimilarity vectors, shape = (rdms, conds)
+        method (str): one of 'evidence', 'setsize' or 'simple'.
+
+    Returns:
+        (ndarray, ndarray): Tuple of the aligned dissimilarity vectors
+            and the weights used
+    """
+    n_rdms, n_conds = dissim.shape
+    if method == 'evidence':
+        weights = (dissim ** 2).clip(0.2)
```
HeikoSchuett: a hardcoded limit on the dissimilarities values seems dangerous to me

HeikoSchuett:
OK, so overall this looks quite good, but I am a bit uncertain about what the align function actually does. It seems to do some rescaling of the dissimilarities, but I don't really understand why. Could you explain that a bit?

iancharest:
@HeikoSchuett a few points regarding your questions below. 

I guess one thing that perhaps isn't clear enough currently is that this is particular for the aligning of behavioural RDMs collected during the multiple arrangements (MA) task. The procedure Jasper implemented here is the iterative weighted averaging that is used to scale and align arrangements of items that are dropped in the multiple trials of the task. the 0.2 hard coded limit is following the heuristic that the the evidence weight of a given distance is SNR^2 (optimal filter theorem). For two diametrically opposed objects at opposite sides of the arena, the distance is 1 and the evidence weight is one. for a distance of .5, the SNR is .5, and the evidence weight is .25. thus it takes four of those latter distances to add up to the same evidence as one of the former. however, for small distances within the range of the placement noise, this evidence weighting rule does not make sense: a distance of 0.1 would yield an evidence weight of 0.01, requiring 100 trials to match the maximum single-trial evidence weight of 1. for closely placed objects, the estimate of the evidence weight itself is unreliable because the placement noise dominates. so here we use the heuristic that distances below 0.2 (about one fifth of the diameter) should all have the same minimum evidence weight of 0.2^2=0.04, requiring about 25 trials to add up to the maximum single-trial evidence weight of 1. 

But perhaps what became clear from your comments is that this align case is quite specific to the multiple arrangements task, so I can think of three options for this pull request:

1- we rename the function to ma_align and make it explicit in the doc that this is a particular case for iterative weighted averaging for the MA task. 

2- we have a meadows module that lives within pyrsa
```python
from pyrsa.meadows import io as mio
from pyrsa.meadows import align
```
3- we move the meadows functionalities outside of pyrsa, which I am not entirely sure is best. we have increasing demands for tools that would enable researchers using the multiple arrangements task to analyse their data themselves, so a module in pyrsa might also have the added benefit that users could use pyrsa for inference too.

HeikoSchuett:
@iancharest 
thank you for your comments! That helped me quite a bit to understand the code and its purpose. 

I think having this code in the toolbox is very sensible. I would currently prefer the version with the submodule where this is not a function of the RDM object. All the other functions could be made such that they are useful and reasonably self explanatory for other purposes as well, but align is too common a word I think for something this specific. I think that it would be much clearer to put this function in a submodule specifically for this kind of behavioral data collected with the units you converged to.

HeikoSchuett:
Hi Jasper,
did my comments make sense? Are you planning to refactor this soon?

pyrsa/rdm/align.py
JasperVanDenBosch: Here's the original comment from the matlab code:

> % we assume constant placement noise. therefore the dissimilarity SNR is
> % proportional to the placement distance in any particular arrangement.
> % the evidence weight of a given distance is SNR^2 (optimal filter
> % theorem). for two diametrically opposed objects at opposite sides of
> % the arena, the distance is 1 and the evidence weight is one. for a 
> % distance of .5, the SNR is .5, and the evidence weight is .25. thus it
> % takes four of those latter distances to add up to the same evidence 
> % as one of the former.
> % however, for small distances within the range of the placement noise,
> % this evidence weighting rule does not make sense: a distance of 0.1 would
> % yield an evidence weight of 0.01, requiring 100 trials to match the
> % maximum single-trial evidence weight of 1. for closely placed objects,
> % the estimate of the evidence weight itself is unreliable because the
> % placement noise dominates. so here we use the heuristic that distances
> % below 0.2 (about one fifth of the diameter) should all have the same minimum
> % evidence weight of 0.2^2=0.04, requiring about 25 trials to add up to the
> % maximum single-trial evidence weight of 1.

JasperVanDenBosch:
> Hi Jasper,
> did my comments make sense? Are you planning to refactor this soon?

To be honest I haven't really figured out an elegant way to deal with this. It's quite a common use case for the behavioral data and I've been trying to get them (people I collaborate with on such projects) to start using pyRSA.

#### A few somewhat incoherent thoughts

1. note that that there are three methods implemented in align for weighting the evidence, for those users that do not want to apply the set minimum (the 0.2**2 value is indeed specific for cases with a distance between 0 and 1, and noise characteristics, so does not apply to all distance metrics). 
2. I'm definitely open to other naming. The matlab name is `estimateDissimMatFromStackOfPartials` which I find a bit wordy so I looked for the verb that was used most in the code to describe it and came up with `align`. converge? rescale? 
3. I did not come up with this algorithm it's a port of the matlab code (kriegeskorte & mur 2012). I just refactored it but the included unit and acceptance tests ensure that the outcomes are identical up to 6 decimal points (which is when scipy imported matlab floats tend to fail anyway)
4. Are we sure that this procedure is not required for other use cases? The main aim is to get around values that are missing among repeated measurements of a set of conditions. 

I'm not fond of the idea of putting it in a `meadows` module as this code it's certainly not meadows-specific. Maybe it's silly to nitpick about naming but I think it makes all the difference for end-users' experience with the library. But otherwise? `BehavioralRDM`.. seems to be making things too complicated.

HeikoSchuett:
sorry to keep you waiting!

You are right that most of these functions are quite general and could be used for other purposes (essentially everything except for the "align" function).
I am a bit worried still that this might be confusing to anyone who has not encountered the multiarrangement task, as I was confused by the naming searching for some dimensions or other RDMs the object was aligned to. To solve this, what do you think about the following division:

- RDMs keeps only the `mean` function with my additions, i.e. by default no weighting and if you pass a weights vector or descriptor name this allows a weighted mean.
- We keep `expand` as a function in pyrsa.rdm renamed as something like `from_list` or `combine_partials` (and add a bit of help that they take a list of partial rdms as input and return a combined RDMs object where each RDM has all dissimilarities)
- provide `align` in pyrsa.rdm as well (probably renamed. I like `rescale` or something like `rescale_combine`) which directly does the whole transformation from a list of partial RDMs to the one weighted average overall RDM. 

With this split we would have `mean` and `expand` available as functions, which are often useful without any idiosyncracies which are only relevant for this rescaling of the partials. And align would be more convenient I think, as I cannot think of a use case for align where we don't want to expand partial RDMs before and average after the align step.

What do you think?

JasperVanDenBosch:
Hi @HeikoSchuett your comments made a lot of sense so I've now tried to restructure it accordingly, what do you think?
I've also added type annotations, see #177, but I can remove them again if we decide to keep python 3.6 for now.

HeikoSchuett (approved):
looks good now I think

PR #140, opened 2020-10-06
Store descriptors as arrays (instead of as lists) in RDMs

charlesbmi:
Summary: Store descriptors as arrays (instead of as lists) in RDMs, so they can be indexed into by the `rdms.reorder` method.

I noticed that `calc_rdm` converts conditions that have `numpy` array values into lists. This appears to happen within `pyrsa.data.computations`:`average_dataset_by`. Seems better to keep everything as a numpy array instead of as a list.

Steps to replicate:
```python
import pyrsa
import numpy as np
import string

nChannel = 50
nObs = 10

# Choose some random letters as our 'conditions'
conditions = np.array(['a', 'a', 'b', 'b', 'c', 'c', 'd', 'd', 'e', 'e'])

obs_des = {'condition' : conditions}
measurements = np.random.rand(nObs, nChannel)

# Create a pyrsa dataset
data = pyrsa.data.Dataset(measurements,
                          obs_descriptors=obs_des)
# By default, the RDM will sort the conditions
rdms = pyrsa.rdm.calc_rdm(data, method='crossnobis', descriptor='condition')

# But perhaps I want the display order to be different
new_order = np.array([0, 4, 2, 1, 3])
rdms.reorder(new_order)

print(rdms)
```
Current output:
```python
  File "quick_test.py", line 22, in <module>
    rdms.reorder(new_order)
  File "/home/charles/Development/pyrsa/pyrsa/rdm/rdms.py", line 354, in reorder
    self.pattern_descriptors[dname] = descriptors[new_order]
TypeError: only integer scalar arrays can be converted to a scalar index
```
Expected output (aside from randomness in `dissimilarities`):
```
pyrsa.rdm.RDMs
1 RDM(s) over 5 conditions

dissimilarity_measure = 
crossnobis

dissimilarities[0] = 
[[ 0.         -2.08915991 -0.25144165 -1.35402033  0.08449471]
 [-2.08915991  0.          0.77741299 -1.80890735 -1.52122028]
 [-0.25144165  0.77741299  0.         -1.2876551  -2.62451962]
 [-1.35402033 -1.80890735 -1.2876551   0.         -1.27272867]
 [ 0.08449471 -1.52122028 -2.62451962 -1.27272867  0.        ]]

descriptors: 
noise = None
cv_descriptor = cv_desc

rdm_descriptors: 
index = [0]

pattern_descriptors: 
index = [0 4 2 1 3]
condition = ['a' 'e' 'c' 'b' 'd']
```

(Python 3.8, using latest `pyrsa` `master` branch)
Still new to this library, so let me know if this is incorrect usage.

### DIFF
--- a/pyrsa/data/computations.py
+++ b/pyrsa/data/computations.py
@@ -35,4 +35,4 @@ def average_dataset_by(dataset, by):
     descriptor = [d.obs_descriptors[by][0] for d in datasets]
     average = [average_dataset(d) for d in datasets]
     n_obs = [d.n_obs for d in datasets]
-    return np.array(average), descriptor, n_obs
+    return np.array(average), np.array(descriptor), np.array(n_obs)


JasperVanDenBosch:
This definitely looks like a bug in `reorder`. Generally speaking I'm not a fan of storing strings etc in a numpy array, I've seen that it can lead to unexpected side effects (all items in the array must conform to the same type meaning the type (string encoding) is implicitly changed).
I also imagine this change would have effects in other places in the toolbox. On other fix would be to change `reorder` to order the descriptor as a list (e.g. `out = [in[i] for i in new_order]`).
In any case we do have to come up with a general rule about whether the descriptors are a list or an array I think.

charlesbmi:
@ilogue that's a good point about string-encoding I hadn't thought about, so I'm good with either lists or arrays. Opened an alternative PR with the `out = [in[i] for i in new_order]` solution you mentioned, depending on what you prefer! https://github.com/rsagroup/pyrsa/pull/141

JasperVanDenBosch:
> @ilogue that's a good point about string-encoding I hadn't thought about, so I'm good with either lists or arrays. Opened an alternative PR with the `out = [in[i] for i in new_order]` solution you mentioned, depending on what you prefer! #141

Well it's not just up to me :) I have some preference for the list- format, so I've approved the other PR. Would be great if someone else could also weigh in.

PR #141, opened 2020-10-07
Update rdms.reorder to assume descriptors is stored as a list

charlesbmi:
An alternate solution to fix the indexing problem in https://github.com/rsagroup/pyrsa/pull/140 . Some discussion there on whether it would be better to store `descriptors` text as a `list` or `np.array`.

Steps to replicate:
```python
import pyrsa
import numpy as np
import string

nChannel = 50
nObs = 10

# Choose some random letters as our 'conditions'
conditions = np.array(['a', 'a', 'b', 'b', 'c', 'c', 'd', 'd', 'e', 'e'])

obs_des = {'condition' : conditions}
measurements = np.random.rand(nObs, nChannel)

# Create a pyrsa dataset
data = pyrsa.data.Dataset(measurements,
                          obs_descriptors=obs_des)
# By default, the RDM will sort the conditions
rdms = pyrsa.rdm.calc_rdm(data, method='crossnobis', descriptor='condition')

# But perhaps I want the display order to be different
new_order = np.array([0, 4, 2, 1, 3])
rdms.reorder(new_order)

print(rdms)
```
Current output:
```python
  File "quick_test.py", line 22, in <module>
    rdms.reorder(new_order)
  File "/home/charles/Development/pyrsa/pyrsa/rdm/rdms.py", line 354, in reorder
    self.pattern_descriptors[dname] = descriptors[new_order]
TypeError: only integer scalar arrays can be converted to a scalar index
```
Expected output (aside from randomness in `dissimilarities`):
```
pyrsa.rdm.RDMs
1 RDM(s) over 5 conditions

dissimilarity_measure = 
crossnobis

dissimilarities[0] = 
[[ 0.         -2.08915991 -0.25144165 -1.35402033  0.08449471]
 [-2.08915991  0.          0.77741299 -1.80890735 -1.52122028]
 [-0.25144165  0.77741299  0.         -1.2876551  -2.62451962]
 [-1.35402033 -1.80890735 -1.2876551   0.         -1.27272867]
 [ 0.08449471 -1.52122028 -2.62451962 -1.27272867  0.        ]]

descriptors: 
noise = None
cv_descriptor = cv_desc

rdm_descriptors: 
index = [0]

pattern_descriptors: 
index = [0 4 2 1 3]
condition = ['a' 'e' 'c' 'b' 'd']
```

(Python 3.8, using latest `pyrsa` `master` branch)

### DIFF
--- a/pyrsa/rdm/rdms.py
+++ b/pyrsa/rdm/rdms.py
@@ -186,7 +186,7 @@ class RDMs:
         Args:
             by(String): the descriptor by which the subset selection
                         is made from descriptors
-            value:      the value by which the subset selection is made
+            value:      the value(s) by which the subset selection is made
                         from descriptors
 
         Returns:
@@ -195,21 +195,18 @@ class RDMs:
         """
         if by is None:
             by = 'index'
+        desc = self.pattern_descriptors[by]  # desc is list-like
         if (
-                type(value) is list or
-                type(value) is tuple or
-                type(value) is np.ndarray):
-            desc = self.pattern_descriptors[by]
-            selection = [np.asarray(desc == i).nonzero()[0]
-                         for i in value]
-            selection = np.concatenate(selection)
+                isinstance(value, list) or
+                isinstance(value, tuple) or
+                isinstance(value, np.ndarray)):
+            selection = np.array([index for index, d in enumerate(desc) if d in value])
         else:
-            selection = np.where(self.rdm_descriptors[by] == value)
+            selection = np.array([index for index, d in enumerate(desc) if d == value])
         selection = np.sort(selection)
         dissimilarities = self.get_matrices()
         for i_rdm in range(self.n_rdm):
             np.fill_diagonal(dissimilarities[i_rdm], np.nan)
-        selection = np.sort(selection)
         dissimilarities = dissimilarities[:, selection][:, :, selection]
         descriptors = self.descriptors
         pattern_descriptors = extract_dict(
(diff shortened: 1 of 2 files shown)


JasperVanDenBosch (approved):
This looks good if we decide to stick with lists for descriptors

JasperVanDenBosch:
I wonder why the test did not pick up on this

HeikoSchuett:
So I just had a look at this and agree with @ilogue. Descriptors should be lists, because there are a couple of objects like strings or arrays of varying size, which do not translate well into arrays. 

However, this seems to be a bit of a mess at the moment as we mostly allow arrays so far and actually convert the descriptors into arrays if the constructor is used. This is also the reason why the test did not pick up on this as I found out. Thus, I think it would be great to at least change this conversion and tidy up anything that breaks once we change this. 

The conversion happens here:
```
def check_descriptor_length(descriptor, n_element):
    """
    Checks whether the entries of a descriptor dictionary have the right length

    Args:
        descriptor(dict): the descriptor dictionary
        n_element: the correct length of the descriptors

    Returns:
        bool

    """
    for k, v in descriptor.items():
        **v = np.asarray(v)**
        if not v.shape:
            # 0-d array happens e.g. when casting str to array
            v = v.flatten()
        descriptor[k] = v
        if v.shape[0] != n_element:
            return False
    return True
```

Do you want to give it a try?

charlesbmi:
yes, I can give it a try this weekend. thanks for the guidance!

JasperVanDenBosch:


```diff

Issues
======
+ Solved 6
           

Complexity increasing per file
==============================
- pyrsa/util/descriptor_utils.py  1
- pyrsa/rdm/rdms.py  2
         

Clones added
============
- pyrsa/rdm/rdms.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/rsagroup/pyrsa/pullRequest?prid=6295791&bid=20479260)

charlesbmi:
I noticed that the `Dataset` object also assumes array-like descriptors, too. Will this change of convention will confuse people who are already using pyrsa?

In a completely separate direction, perhaps these descriptors would be a good place to use DataFrames, instead of dictionaries of arrays/lists? DataFrames would combine the ease of array-indexing while overcoming the mixed-type issue.

HeikoSchuett:
Hi there!
sorry for letting this one fall asleep. 
It looks like this currently breaks many tests as the generation of RDM objects now fails?  I would like to get to the bottom of this now. Are you planning to continue working on this? Otherwise I would take your code from here and go through this to fix this issue once and for all.

Otherwise on the Dataset object: Yes, I would also also want to change this in the Dataset object

charlesbmi:
Hi Heiko,

I didn't get around to this one. Feel free to take it up!

Best,
Charles

PR #155, opened 2020-10-19
Implementing CodeFactor suggestions

HeikoSchuett:
I'll have a look at the suggestions of the continuous integration machines to improve the code quality a bit.
Especially CodeFactor thinks our code is really bad.


JasperVanDenBosch:
I wouldn't worry tooo much about codefactor, it doesn't seem to be a very mature product yet. It doesn't seem to take our config into account well and gets confused with PR branches it seems..

HeikoSchuett:
I will keep this in mind! Thanks for the comment! There are some issues a few of the code style checks respond to though, so starting a pull request where we/I can go and fix some of those to make nicer code, when we have some time seems like a good idea nonetheless I think.

tests/test_rdm.py
```diff
@@ -501,166 +444,6 @@ def test_calc_rdm_movie_binned(self):
         assert rdm.rdm_descriptors['time'][0] == np.mean(time[:3])
 
 
-class TestCompareRDM(unittest.TestCase):
```
JasperVanDenBosch: @HeikoSchuett I think something went wrong here during a merge; large parts of test_rdm are not on this branch anymore; the commit is probably 75633df0741392b3f0445b172619260cc34808c7

Would it make more sense to try to fix this branch or start over from a clean branch? You'd remember best if a lot of effort went into this already.

HeikoSchuett:
I am closing this one as it is superseded by our new code checks.

PR #158, opened 2020-10-21
Rdm plot updates

jooh:
I'm working on the updates from the hackathon. So far we have

* image labels in groups (by adding Icon instances as a pattern descriptor)
* a minimal demo notebook (demos/exercise_rdm_visualisation.ipynb)

Right now I'm working on
* auto-setting zoom and offset for the images by playing with `transAxes.transform`. For now you have to tweak these by hand
* Only plotting image labels along the marginal panels in subplots

There are many other things to look at in the Hackathon doc. We will see if I get to all of them. Give me a shout if you are also working on `show_rdm` so we can avoid duplicated effort.

rdm


JasperVanDenBosch:


```diff

Issues
======
- Added 2
           

Clones removed
==============
+ pyrsa/vis/icon.py  -8
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/rsagroup/pyrsa/pullRequest?prid=6378838&bid=20692379)

jooh:
Ok, this is ready now I think. Key changes, mainly concerning the main API at `vis.show_rdm`:
* Image labels! With auto-scaling to control overlap!
* Limitation of the above - works well with inline notebook back end and for saving (with `bbox_inches='tight'`), but not so well for classic floating figure windows. In general the scaling is not dynamic, ie, if the figure changes size all bets are off.
* Multi-panel figures have labels along the marginal axes only
* Colorbars are per RDM or per figure. If it's per figure we set a shared `vmin` and `vmax` so the scale is the same over panels.
* The diagonal is masked out by default. You can mask other things as desired by passing `nanmask`.
* We use a new style sheet `rdm.mplstyle` to control various figure parameters concisely. We should do this more in other visualisations to achieve a consistent toolbox appearance.
* The `do_ranktransform` argument is no more, because it messes with the color bars. If you want to rank transform it's better to do so before visualising with `rdm.rank_transform` (and changing `dissimilarity_measure` accordingly).
* Many edits to methods in `vis` and `rdm_plot` to be more sensible about using Matplotlib's OO API instead of relying on `plt`, and to be nice about returning handles to graphics objects to support this.
* Broke out sub-parts of the `show_rdm` functionality into functions that can be called independently: `show_rdm_panel` for drawing into an existing axis, and `add_descriptor_x_labels` / `add_descriptor_y_labels` to add fancy image labels to completely different things (you could probably use this to put labels on model comparison graphs if you consider each RDM an Icon instance...).
* Implemented type hints. This will be *a lot* more useful when Matplotlib gets around to implementing them.
* Ran Black on the modules I worked on. They look better now.
* New notebook demonstrating everything: `demos/rdm_visualisation_92images.ipynb`

And here's what it looks like:
temp_rdm

JasperVanDenBosch (approved):
Thanks Johan!

wenx-guo:
@jooh 
Hi Johan, thank you for adding this nice functionality. However, there seems to be some spacing issues here. I tried out the demo demo_rdm_visualisation_92images.ipynb. 

Here are the figures generated using the demo code:
image
image

Using plt.tight_layout() or plt.gcf().subplots_adjust doesn't work well -- the image labels are always cut off.

jooh:
@WxComeon Thanks for the report. If you save the figure using the lines in the demo notebook, does the resulting figure still have cropped labels? Generally this happens if you do not have `bbox_inches='tight'` set, perhaps because you have changed your notebook defaults.

https://rsatoolbox.readthedocs.io/en/latest/demo_rdm_vis.html#Optimising-image-label-display-and-saving-figure

wenx-guo:
@jooh Thanks! I did try with saving this figure with bbox_inches set to 'tight'. Unfortunately the figure still looks like the above. I made sure I have the updated version of rsatoolbox, and I don't think I changed my notebook defaults on this setting. 

I also tried with adding plt.tight_layout(), which cut the icons off entirely (comparing to the first figure I posted above). I think the icons are added as tick labels, so plt.tight_layout() should potentially take care of it. Somehow it doesn't work. I made some casual plots with large axis labels, and tight_layout works.
image

JasperVanDenBosch:
For me changing the size of the figure helped. But I don't know how that translates to jupyter

jooh:
Hm. @WxComeon do you have the ability to test the code in a clean environment, for instance by running the notebook in a docker container? It might help us get to the bottom of why this happens. It works on my mac, and it worked in the docs build context. It would be good to have a reproducible environment where it doesn't work to debug.

In general - a weakness of the current code is that the labels are placed outside the bounds of the nominal figure. `bbox_inches='tight'` works because it basically sets the printed figure size to encompass all the plot objects, so the figure expands to fit effectively.

A more robust approach would be to try to position the subplot grid to allow space on the margin for the label. But this is a little hairy with matplotlib's imperative API. You would basically need a first pass to just work out where the labels are going to be, then adjust figure size and gridspec to accommodate them, then re-do all the label calculations (since they won't be valid anymore after changing the figure size), then finally drawing everything. Could be a fun PR for someone, but I don't have time sadly!

wenx-guo:
@jooh @ilogue Thanks so much for the suggestions. The problems indeed stemmed from my settings of the notebook. I haven't found where the problems are yet, but it does work well in a docker container. I agree that calculating the ratios and sizes is really tricky here...

PR #159, opened 2020-10-28
Demo tests

### LINKED ISSUE #142: Testing demo notebooks and pytest
It would be useful to include the demo notebooks in tests to detect any regressions. I did some research and there seems to be two ways you could do this.

1. Add dependencies: pytest and [nbval](https://nbval.readthedocs.io/en/latest/). This is the most elegant approach. See e.g. [this example application](https://github.com/miykael/workshop_cambridge/blob/master/test_notebooks.py). But we would be switching the test backend to pytest (which is compatible with the existing unittest tests, but all the same, it will at a minimum require messing with CI). For what it's worth, I think pytest is much nicer than unittest anyway.

2. You can hack together your own unittest notebook runner by running something like `jupyter nbconvert --to notebook --inplace --execute mynotebook.ipynb` in a subprocess, and re-throwing any exceptions that come back from the shell. I got this idea from [this SO answer](https://stackoverflow.com/a/40311709/3375155). No further dependencies, but I have a feeling this may or may not work well in practice, since you're calling a nested python process and that introduces fun possibilities like somehow picking up a *different* python from the original test runner...

Any thoughts on this?

HeikoSchuett:
This pullrequest adds tests which run the same code as the demos. This should check that the demos can still run and also serves as a more overarching test that things we actually want to do still run. 

While creating these I also fixed a few smallish issues.


JasperVanDenBosch:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- tests/test_demo.py  9
         

Clones added
============
- tests/test_demo.py  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/rsagroup/pyrsa/pullRequest?prid=6419245&bid=20796649)

JasperVanDenBosch (approved):
The downside of course is that any change to the notebooks would also have to be made in these tests.. I hope people will have the discipline for that :stuck_out_tongue: 

But in the mean time it's a good compromise

iancharest (approved):
This looks good. quick suggestion, should we have a 
```python
%matplotlib inline
```
at the top of our jupyter notebooks?

PR #163, opened 2020-11-10
nan handling for _cov comparisons

HeikoSchuett:
The current implementation of the _cov measures (with jörns whitening) breaks with nan values. This pull request fixes this issue.

There are still bonus points to be earned for fixing the faster linear_CKA version


adkipnis (approved):
Looks good and seems to work in practice (albeit with substantially decreased speed)

PR #167, opened 2020-12-16
Add option to reorder conditions using an explicit list

charlesbmi:
`pyrsa` currently orders RDM conditions by order of appearance. At times, the conditions have some non-alphabetical relation to each other that we would like to highlight. For example, the order of digits on the hand is not alphabetical by name.

This change allows the `pyrsa` user to explicitly pass in the order of conditions.

Example usage (in `pdb` prompt):
```python
-> data_rdm.sort_by(condition=condition_order)
(Pdb++) data_rdm  # before .sort_by
pyrsa.rdm.RDMs(
dissimilarity_measure = 
correlation
dissimilarities = 
[[0.04313705 0.02874738 0.03569133 0.05490419 0.04283459 0.0449416
  0.03896742 0.02868705 0.04261004 0.04269907]]
pattern_descriptors = 
{'index': array([0, 1, 2, 3, 4]), 'condition': array(['R', 'T', 'M', 'I', 'P'], dtype='<U1')}

(Pdb++) condition_order  # desired new order
['T', 'I', 'M', 'R', 'P']
(Pdb++) n  # run the sort_by line
(Pdb++) data_rdm  # after .sort_by
pyrsa.rdm.RDMs(
dissimilarity_measure = 
correlation
dissimilarities = 
[[0.0449416  0.04283459 0.04313705 0.03896742 0.02868705 0.03569133
  0.04269907 0.02874738 0.04261004 0.05490419]]
pattern_descriptors = 
{'index': [1, 3, 2, 0, 4], 'condition': ['T', 'I', 'M', 'R', 'P']}
```

(This specific example of hand digits would be the precursor to creating a figure like https://www-nature-com.clsproxy.library.caltech.edu/articles/nn.4038/figures/3 where the labels say `['T', 'I', 'M', 'R', 'P']` instead of numbers.


JasperVanDenBosch (changes requested):
This looks good, but it would be great if you could add a  unit test

Have a look here https://github.com/rsagroup/pyrsa/blob/master/tests/test_rdm.py#L299 for an example.
Ping me if I can help

pyrsa/rdm/rdms.py
```diff
@@ -357,7 +357,9 @@ def sort_by(self, **kwargs):
         """Reorder the patterns by sorting a descriptor
 
         Pass keyword arguments that correspond to descriptors,
-        with value 'alpha'. 
+        with value indicating the sort type. Supported methods:
+            'alpha': sort alphabetically (using np.sort)
+            list/np.array: specify the new order explicitly
```
JasperVanDenBosch: also maybe clarify that the values in the list/array should correspond to the descriptor values; as opposed to just the indices that we'd pass to `reorder()`

charlesbmi:
@ilogue I added the unit test but am not sure what to do about the "Method could be a function" Codacy error - isn't this error applicable to most unit tests?

HeikoSchuett:
Hi @charlesincharge !
Yes, Codacy is very thorough in this case and we have quite a few tests which violate this rule. This is not really a problem and if it is hard to fix we just ignore Codacy.
Here, this can be fixed relatively easily by replacing the assert [...] statement with a self.assert[...] statement though, which is the preferred formulation for the unittest library anyway I think.

JasperVanDenBosch (approved):
Good stuff!

HeikoSchuett (approved):
Looks good overall

charlesbmi:
Thanks!

PR #168, opened 2021-01-14
Normalize crossnobis distance by number of channels and fold-iterations

charlesbmi:
To match the documentation,
image
the crossnobis sum should be divided by the number of channels `P` and the number of iterations `M-choose-2`.

Note: `_calc_crossnobis_single`, which is used if a list of noises is provided, does already normalize by the number of channels: https://github.com/rsagroup/pyrsa/blob/master/pyrsa/rdm/calc.py#L443

This change causes the crossnobis distance to match the scaling of mahalanobis distance.

### DIFF
--- a/pyrsa/rdm/calc.py
+++ b/pyrsa/rdm/calc.py
@@ -297,7 +297,7 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
     dataset.sort_by(descriptor)
     cv_folds = np.unique(np.array(dataset.obs_descriptors[cv_descriptor]))
     rdms = []
-    if noise is None or (isinstance(noise, np.ndarray) and noise.ndim == 2):
+    if (noise is None) or (isinstance(noise, np.ndarray) and noise.ndim == 2):
         for i_fold in range(len(cv_folds)):
             fold = cv_folds[i_fold]
             data_test = dataset.subset_obs(cv_descriptor, fold)
@@ -307,7 +307,7 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
                 average_dataset_by(data_train, descriptor)
             measurements_test, _, _ = \
                 average_dataset_by(data_test, descriptor)
-            n_cond = measurements_train.shape[0]
+            n_cond, n_channel = measurements_train.shape
             rdm = np.empty(int(n_cond * (n_cond-1) / 2))
             k = 0
             for i_cond in range(n_cond - 1):
@@ -339,7 +339,7 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
                                       + variances[j_fold]))
                     rdms.append(rdm)
     rdms = np.array(rdms)
-    rdm = np.einsum('ij->j', rdms) / len(cv_folds)
+    rdm = rdms.mean(axis=0)
     rdm = RDMs(dissimilarities=np.array([rdm]),
                dissimilarity_measure='crossnobis',
                rdm_descriptors=deepcopy(dataset.descriptors))
--- a/pyrsa/rdm/calc.py
+++ b/pyrsa/rdm/calc.py
@@ -308,7 +308,6 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
             measurements_test, _, _ = \
                 average_dataset_by(data_test, descriptor)
             n_cond = measurements_train.shape[0]
-            n_channel = measurements_train.shape[1]
             rdm = np.empty(int(n_cond * (n_cond-1) / 2))
             k = 0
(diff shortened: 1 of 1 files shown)


charlesbmi:
fyi @HeikoSchuett @ilogue . ~~I'm not sure if this is already handled by averaging somewhere else in the function?~~

pyrsa/rdm/calc.py
```diff
@@ -314,18 +315,15 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
...
-            variances.append(np.linalg.inv(noise[i_fold]))
         for i_fold in range(len(cv_folds)):
-            for j_fold in range(i_fold + 1, len(cv_folds)):
+            for j_fold in range(len(cv_folds)):
                 if i_fold != j_fold:
                     rdm = _calc_rdm_crossnobis_single(
                         measurements[i_fold], measurements[j_fold],
-                        np.linalg.inv(variances[i_fold]
-                                      + variances[j_fold]))
+                        noise[i_fold])
```
charlesbmi: Made this change because adding 2 variances resulted in reporting each distance as half as large.

HeikoSchuett: Hi!
Thanks for placing a pointer to this location.

 Here I fear we need this combination of the two variances.
This is the case where we have different noise variances for the different folds. This combination of the two variances leads to a down weighting whenever one of the two variances is large, i.e. for all combinations where we are unsure about either measurement. After your change we would care only about the noise in one of the two folds (the i_fold).  

Overall scaling should be fine nonetheless. Thus, I would suggest to put the inverse of the sum back in but then multiply the overall result by two.

HeikoSchuett:
Thanks for tackling this consistency!

pyrsa/rdm/calc.py
charlesbmi: How about if we iterate both `i_fold`,`j_fold` over all fold permutations (changes at line 322)?

I think the math would be equivalent (?) but it would follow the crossnobis equation in `Walther et al. (Neuroimage 2016)` a little more obviously:
image
(which only uses a single `noise[fold_A]` per split)

HeikoSchuett: Unfortunately the math is not equivalent. What you suggest would be:
<html>
<body>
d<sup>2</sup> = (b<sub>j</sub>-b<sub>k</sub>)<sub>A</sub> S<sub>A</sub><sup>-1</sup> (b<sub>j</sub>-b<sub>k</sub>)<sub>B</sub> <sup>T</sup> + (b<sub>j</sub>-b<sub>k</sub>)<sub>A</sub> S<sub>B</sub><sup>-1</sup> (b<sub>j</sub>-b<sub>k</sub>)<sub>B</sub> <sup>T</sup> <br>
= (b<sub>j</sub>-b<sub>k</sub>)<sub>A</sub>( S<sub>A</sub><sup>-1</sup> + S<sub>B</sub><sup>-1</sup> ) (b<sub>j</sub>-b<sub>k</sub>)<sub>B</sub> <sup>T</sup> <br>
&ne;(b<sub>j</sub>-b<sub>k</sub>)<sub>A</sub>( S<sub>A</sub> + S<sub>B</sub> )<sup>-1</sup> (b<sub>j</sub>-b<sub>k</sub>)<sub>B</sub> <sup>T</sup> <br>
</body>
</html>

where the last unequal line would be what the code did before the change. 

Origninally, I implemented a switch to choose how the variances for the two folds should be included, but this made the code here quite messy and I thus decided to reduce it to one variant. This is intentionally slightly different from the Walther et al. formulation. The main reason for this is what I alluded to in my earlier comment: In the Walther et. al formula the summand described will still have a strong influence even if the variance in B was gigantic, which we want to avoid. Thus, I would like to keep the original formulas arguing that they are a better formula than what Walther proposed in this fringe case.

tldr: this is not the same and the difference to Walther et al. is intentional, please keep it.

charlesbmi: Understood, thanks for the explanation! I'll update this PR.

charlesbmi: Updated! Looks like Codacy is failing because it isn't tracking the `main` branch? https://app.codacy.com/gh/rsagroup/pyrsa/pullRequest?prid=6800656 @HeikoSchuett

charlesbmi:
Going to try rebasing and seeing if this fixes the codacy error

HeikoSchuett (approved):
seems now to do the right thing.

charlesbmi:
Thanks! Would someone else be able to give me a 2nd review? Maybe @Benjamin-Peters or @ilogue ?

JasperVanDenBosch (approved):
LGTM

PR #171, opened 2021-01-23
Ignore PyCharm's local config files

caiw:


caiw:
Simple one: this is PyCharm IDE's local config

JasperVanDenBosch (approved):

PR #172, opened 2021-01-23
concat can accept RDMs as separate arguments

### LINKED ISSUE #170: `concat(rdms)` should be able to accept multiple RDMs as arguments
It would be nice (and expected?) to be able to write
```python
rdms = concat(rdms1, rdms2, rdms3)
```
instead of 
```python
rdms = concat((rdms1, rdms2, rdms3))
```
This can be easily achieved by changing the signature of `concat` to `concat(*rdms)`, and doing a little check to ensure lists can still be passed.

caiw:
Addresses Issue  #170: `concat` can accept either a `list` (or other iterable) of `RDMs`s, or can accept individual `RDMs`s as arguments.


HeikoSchuett (approved):
Does what it should do

HeikoSchuett (changes requested):
Oh, I just tried this out. There seems to be a problem with the tests at least. Due to the reference character of the concat we get running concat twice does something interesting here. It seems the output of the two concats you compare in the test in the end actually link to the same object which then has more than 16 entries. 
I noticed this, because  == should not work properly for RDMs objects so far

A possible solution would be to do a copy of the first RDM at the beginning before appending RDMs.

caiw:
Ok, I'll have a look at this.  If `==` is doing something unexpected I'll open a separate issue to address that.

HeikoSchuett:
It does not do anything unexpected, but it checks that the two objects are exactly the same, i.e. here the output creates the exact same object

HeikoSchuett (approved):
now everything should be fine

PR #176, opened 2021-02-12
Change above-0 test to use `<=`

charlesbmi:
I'm not sure it matters much, but I think the null hypothesis for `test > 0` is `<=0`


HeikoSchuett (approved):

HeikoSchuett:
Sure, exact equality should hardly ever happen, but including equality here is probably more correct.

charlesbmi:
Thanks! Would someone else be able to give me a 2nd review? Maybe @ilogue ?

PR #182, opened 2021-03-25
speed update for calc_rdm

### LINKED ISSUE #181: Vectorized distance measures?
In the case of numerous channels (e.g., as in some neural network representations), the current implementation of the RDM calculations requires an immense amount of memory and compute compared to the optimal, vectorized NumPy solution.

For example, for the (squared) Eucledian distance with n = 1000, m = 10000
```
non-vectroized computation time: 21.692641s (pyrsa.rdm.calc_rdm_euclid)
vectorized computation time: 0.075766s
```

Have you guys considered and decided against vectorizing these distance measures?

HeikoSchuett:
As @Tal-Golan  noticed, we had a large performance problem for calculating RDMs in the first place.
The vectorized solution which calculates the kernel/inner product first is MUCH faster. I switched all calc_rdm functions to this method here.

closes #181

### DIFF
--- a/pyrsa/rdm/calc.py
+++ b/pyrsa/rdm/calc.py
@@ -84,9 +84,10 @@ def calc_rdm(dataset, method='euclidean', descriptor=None, noise=None,
     return rdm
 
 
-def calc_rdm_movie(dataset, method='euclidean', descriptor=None, noise=None,
-             cv_descriptor=None, prior_lambda=1, prior_weight=0.1,
-             time_descriptor='time', bins=None):
+def calc_rdm_movie(
+        dataset, method='euclidean', descriptor=None, noise=None,
+        cv_descriptor=None, prior_lambda=1, prior_weight=0.1,
+        time_descriptor='time', bins=None):
     """
     calculates an RDM movie from an input TemporalDataset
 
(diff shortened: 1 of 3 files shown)


pyrsa/rdm/calc.py
```diff
@@ -307,21 +305,8 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
                 average_dataset_by(data_train, descriptor)
             measurements_test, _, _ = \
                 average_dataset_by(data_test, descriptor)
-            n_cond = measurements_train.shape[0]
-            rdm = np.empty(int(n_cond * (n_cond-1) / 2))
```
JasperVanDenBosch, line 311: downside of numpy empty is that its filled with random values, I tend to use `numpy.full(s, numpy.nan)` as it'll be more obvious when something goes wrong

pyrsa/rdm/calc.py
```diff
@@ -519,3 +494,8 @@ def _check_noise(noise, n_channel):
     else:
         raise ValueError('noise(s) must have shape n_channel x n_channel')
     return noise
+
+
+def _extract_triu_(X):
```
JasperVanDenBosch: this may be something we'll eventually need in other modules as well

JasperVanDenBosch (approved):
This is great. I haven't tested it myself but it looks like any discrepancies should be covered by the unit tests.

pyrsa/rdm/calc.py
```diff
@@ -233,14 +231,12 @@ def calc_rdm_mahalanobis(dataset, descriptor=None, noise=None):
...
-        # then calculate the inner product
-        diff = _calc_pairwise_differences(measurements)
-        diff2 = (noise @ diff.T).T
-        rdm = np.einsum('ij,ij->i', diff, diff2) / measurements.shape[1]
+        kernel = measurements @ noise @ measurements.T
+        rdm = np.expand_dims(np.diag(kernel), 0) + np.expand_dims(np.diag(kernel), 1)\
+            - 2 * kernel
+        rdm = _extract_triu_(rdm) / measurements.shape[1]
         rdm = RDMs(dissimilarities=np.array([rdm]),
-                   dissimilarity_measure='Mahalanobis',
```
Tal-Golan, line 243: Shouldn't this remain "dissimilarity_measure='Mahalanobis'"?

HeikoSchuett: yes it should. Is fixed now

pyrsa/rdm/calc.py
HeikoSchuett: moved this into utils!

Tal-Golan (approved):

pyrsa/rdm/calc.py
HeikoSchuett: I often use the same. The empty function is faster though. As I remove this statement here anyway, I just mark this as resolved.

PR #184, opened 2021-03-30
Documentation update for calculating RDMs

### LINKED ISSUE #179: Documentation of distance measures - squared vs non-squared distances
As a user of the toolbox, I find it misleading that 'euclidian distance' is the squared euclidean distance (divided by the number of channels). I suggest changing the documentation of `calc_rdm_euclid`, `calc_rdm_mahalanobis` and `calc_rdm_crossnobis` to reflect the computed distance function more precisely. This will ensure that future users don't incorrectly report the distance measures they've employed.

HeikoSchuett:
This contains an update for the documentation for the calculation of RDMs. 

It shall extend the existing example notebook, the calculations with noise covariances, dealing with non-squared distances and the help messages.


docs/source/Temporal RSA.nblink
```diff
@@ -0,0 +1,6 @@
+{
```
JasperVanDenBosch: Can we replace the space with an underscore or hyphen? Especially since this path will become part of the url. It'd have to be done on the original notebook as well

pyrsa/rdm/rdms.py
```diff
@@ -463,9 +464,58 @@ def rank_transform(rdms, method='average'):
...
-                    descriptors=rdms.descriptors,
-                    rdm_descriptors=rdms.rdm_descriptors,
-                    pattern_descriptors=rdms.pattern_descriptors)
+                    descriptors=deepcopy(rdms.descriptors),
+                    rdm_descriptors=deepcopy(rdms.rdm_descriptors),
+                    pattern_descriptors=deepcopy(rdms.pattern_descriptors))
+    return rdms_new
+
+
+def sqrt_transform(rdms):
```
JasperVanDenBosch: I'd argue for putting these in a separate "transforms" module but that's subjective

JasperVanDenBosch: Also wouldn't this be more attractive as a method?

JasperVanDenBosch (approved):
Excellent work

docs/source/Temporal RSA.nblink
HeikoSchuett: done, actually unified names for all demos

pyrsa/rdm/rdms.py
HeikoSchuett: I am fine with this at least, so I did it

HeikoSchuett: we could use it as a method, too. Would it then make sense to interpret this as an in place operation? If so: Should we then use this as the convention?
`function(object)` returns a new object
`object.function()` changes the object

Didn't want to start such a convention without checking in.

JasperVanDenBosch: Yeah that's a good question. I was thinking to open an issue about this a while ago. I'm fine either way as long as we aim to keep it consistent. Also if possible the methods would return `self` so you can chain them: `rdms.sum().sqrt()`.

Tal-Golan (approved):
I'm sorry about the delay. I haven't gone over all of the commits, but I believe that these edits sufficiently address issue #179.

pyrsa/rdm/rdms.py
HeikoSchuett: I've noticed this applied to a few other transformations of RDMs and data. I will copy this into an issue about this and fix this at all places simultaneously and merge this pull request for now.

PR #193, opened 2021-04-23
Fix for missing mplstyle for rdm_plot

### LINKED ISSUE #191: Get nbsphinx to deal with the custom mplstyle file
This currently fails: https://rsatoolbox--189.org.readthedocs.build/en/189/demo_rdm_vis.html

### LINKED ISSUE #192: show_rdm figures broken in docs due to missing rdms.mplstyle
Looks like vis.show_rdm is broken in the docs because the rdms.mplstyle file is missing from the build context. Happy to fix if someone can point to where it needs to be added in.

See e.g. https://rsatoolbox.readthedocs.io/en/latest/demo_rdm_vis.html#Plotting-RDMs-with-text-labels

JasperVanDenBosch:
fixes #191
fixes #192

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+recursive-include rsatoolbox *.mplstyle
+include requirements.txt
--- a/rsatoolbox/vis/rdm_plot.py
+++ b/rsatoolbox/vis/rdm_plot.py
@@ -5,8 +5,7 @@ Plot showing an RDMs object
 """
 
 from __future__ import annotations
-import os.path
-import inspect
+import pkg_resources
 import collections
 import typing
 from typing import TYPE_CHECKING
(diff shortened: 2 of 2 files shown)


JasperVanDenBosch:
@jooh do you think we can type the `rdm_descriptor` as `str` instead of `Any`?

jooh:
> @jooh do you think we can type the `rdm_descriptor` as `str` instead of `Any`?

I think that depends on the RDMs. Can you set a non-str type as a key for the RDMs.rdm_descriptor dict?

I notice there are more general problems with the logic of how this arg is handled (my bad). If you pass a str (which is *not* a key into rdm_descriptor, this is weird), that same str gets set as the title for *all* RDMs in the multi panel case. If we want to support this kind of multi type case it should probably be passing a list with len == n_rdms.

Or better yet, just enforce only rdm_descriptor keys and force users to set up their RDMs instances properly.

JasperVanDenBosch:
> I think that depends on the RDMs. Can you set a non-str type as a key for the RDMs.rdm_descriptor dict?
> 
> I notice there are more general problems with the logic of how this arg is handled (my bad). If you pass a str (which is _not_ a key into rdm_descriptor, this is weird), that same str gets set as the title for _all_ RDMs in the multi panel case. If we want to support this kind of multi type case it should probably be passing a list with len == n_rdms.
> 
> Or better yet, just enforce only rdm_descriptor keys and force users to set up their RDMs instances properly.

I see! I think we won't want to allow non-string keys (I can't think of a use case) for the descriptors. I think the case where the descriptor does not exist and it falls back to using it as the default title is perfectly fine for now. 
Also I think for cases where we don't know the type yet it's probably fine to just not annotate it until we do (which makes it "Unknown").

HeikoSchuett (approved):

PR #195, opened 2021-04-29
list saving fixed

HeikoSchuett:


JasperVanDenBosch:
Does this occur for string only? Would this fix it: https://stackoverflow.com/a/43935389/708221 ?
I mean, would that be easier than converting to dicts?

HeikoSchuett:
Hm, the pressing problem was that the saving function ignored lists, which was counterproductive since we now have lists instead of arrays as descriptors.

Dealing with lists of strings is a special case, we should cover. I'll have a look whether this works better. I think we still need the dict thing for stuff like the noise covariance matrices, which are not strings

JasperVanDenBosch:
Hm right in that case a more general solution like the dicts may be better indeed

PR #200, opened 2021-06-30
Switch CI to GitHub actions

JasperVanDenBosch:
Since Travis doesn't support free public repositories properly anymore, this switches us to Github Actions for running CI tests


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Update pip
+      run: python -m pip install --upgrade pip
+    - name: Install dependencies
+      run: pip install -r requirements.txt
+    - name: Unit tests
+      run: python setup.py test
```
HeikoSchuett: This looks great overall and seems to work. The examples and I seem to prefer `pytest` over `python setup.py test`, because it produces a nicer overview over the results. Do you have reasons to prefer `python setup.py test`?

If we want to switch to `pytest`,  `python -m pip install pytest` is required with the requirements.

HeikoSchuett (approved):
This all looks great and functional. Consider the switch to pytest

.github/workflows/pull_request.yml
JasperVanDenBosch: I've always preferred the standard lib test runner because there seems to be many third-party ones out there and I didn't want us to start biasing the tests to one tool (eventually making them incompatible with others).
But I've read that `python setup.py test` is going to be deprecated so we do need to find a replacement.
The new default one will be `tox`. I suppose we could do a quick poll to decide which one people prefer?

PR #209, opened 2021-11-11
Meadows json rdms

JasperVanDenBosch:
The `load_rdms` function should also support json/tree files


HeikoSchuett (approved):
This looks fine in general. I am not 100 % on top of the convention of meadows, but assume you are such that this makes sense for all cases you get from there

One question though: Is the petname package really necessary here? It seems like you use these as some kind of default in meadows?

JasperVanDenBosch:
> assume you are such that this makes sense for all cases you get from there

Yes at least for the ones that have pre-made rdms in the data. Next step could be to have functions to load raw pattern data

> One question though: Is the petname package really necessary here? It seems like you use these as some kind of default in meadows?

Yeah I'm not very happy to include it here either but otherwise we'd have to include the list of potential names in a module or something. Unfortunately the way the files are named, this is the only way to distinguish the file types for now..

Issue #211, opened 2021-11-16
If RDM contains NaN, return NaN instead of Error


smazurchuk:
Thanks for making this toolbox. I was wondering if people would be open to changing the behavior of the `_parse_input_rdms` function. Often when doing searchlight analysis, `NaNs` get introduced into some patches. At present the `evaluate_models_searchlight` function raises an error (and stops) if it encounters a NaN. Could we change this behavior to simply add an extra option to `evaluate_searchlight` function to return a NaN vertex value at patches that contain NaNs?

Thanks

https://github.com/rsagroup/rsatoolbox/blob/ead07b256aad5005c883b6fba3cd50c7417bbcf9/rsatoolbox/rdm/compare.py#L520

HeikoSchuett:
apparently @ilogue  will have a look at that. Changing this behavior certainly seems sensible.

I think we might want to keep the check in the `_parse_input_rdms` because it checks whether RDMs that have both NaNs and finite values as entries are compatible. It seems better to me to catch the NaN inducing positions already in the searchlight or compare functions, such that they immediately return `NaN` before starting any computation about the RDMs.

smazurchuk:
Very cool! Thanks for checking this out. I just started using the toolbox. While changing the NaN behavior would be helpful, is there any discussion/interest in improving the performance of the toolbox for searchlight analysis?

I've been unable to get a searchlight analysis down to a reasonable time (even having tried multithreading and multi-process approaches). Linked is a jupyter-notebook I put together to show that I think a key place that could drastically speed up the toolbox is optimizing the `split_obs` function. It is used very often and depends on a list comprehension. I think it might be possible to change this to also accept a numpy boolean vector (or numpy array for fancy indexing)

https://github.com/smazurchuk/general/blob/master/rsa_timing.ipynb

Thanks again! Let me know if this is a direction people are interested in moving.

HeikoSchuett:
I'll have a look at the timing issue. Making the subsetting faster would benefit all analyses in the toolbox, so that is certainly worth a bit of effort.

HeikoSchuett:
Ok, so I had a bit of a look at the timing stuff and think I managed to remove a bit of the baggage which comes from copying around the descriptors of the data everywhere. 

I was not entirely sure how many channels your dataset has, so I set it to 100. With that number of channels I would now be at ~100ms for computing one cross-nobis RDM based on cross validating the 6 runs for 300 patterns, i.e. I ran the following code:

```
[...]

PR #212, opened 2021-11-18
speedup code for computing RDMs

HeikoSchuett:
here specifically skip descriptor things for averaging datasets and save some computation for dataset handling in-between

### DIFF
--- a/rsatoolbox/data/computations.py
+++ b/rsatoolbox/data/computations.py
@@ -54,7 +54,7 @@ def average_dataset_by(dataset, by):
     average = np.nan * np.empty(
         (len(unique_values), dataset.measurements.shape[1]))
     n_obs = np.nan * np.empty(len(unique_values))
-    for i_v, v in enumerate(unique_values):
+    for i_v, _ in enumerate(unique_values):
         measurements = dataset.measurements[inverse == i_v, :]
         average[i_v] = np.mean(measurements, axis=0)
         n_obs[i_v] = measurements.shape[0]
--- a/rsatoolbox/data/computations.py
+++ b/rsatoolbox/data/computations.py
@@ -21,24 +21,6 @@ def average_dataset(dataset):
     return np.mean(dataset.measurements, axis=0)
 
 
-def average_dataset_by_slow(dataset, by):
-    """
-    computes the average of a dataset per value of a descriptor
-
-    Args:
-        dataset(rsatoolbox.data.Dataset): the dataset to operate on
-        by(String): which obs_descriptor to split by
-
-    Returns:
-        numpy.ndarray: average: average activation vector
-    """
-    datasets = dataset.split_obs(by)
-    descriptor = [d.obs_descriptors[by][0] for d in datasets]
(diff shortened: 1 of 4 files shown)


HeikoSchuett:
This is in response to #211

rsatoolbox/data/computations.py
```diff
@@ -20,7 +21,7 @@ def average_dataset(dataset):
     return np.mean(dataset.measurements, axis=0)
 
 
-def average_dataset_by(dataset, by):
+def average_dataset_by_slow(dataset, by):
```
JasperVanDenBosch: I think if the slow one is not referenced anywhere you can just remove it. If we need to see how it used to be implemented we can use git history right?

JasperVanDenBosch (approved):
Looks good to me!

rsatoolbox/data/computations.py
HeikoSchuett: yes, of course, this will be gone when I merge this in

smazurchuk:
Hey @HeikoSchuett , thanks for looking into this. It does seem that this change sped-up the calculation of RDM's, but in a searchlight type analysis I didn't see any net effect. Linked here is a notebook. I was wondering if you could look at it to see if maybe I did something improperly? I'm also not really sure the best way to import a working branch. 

The imported folder is on the proper branch
image

https://github.com/smazurchuk/general/blob/master/speedup_analysis.ipynb

Thanks!

smazurchuk:
Whoops, realized it is not loading the proper file. Is there a way to load both the package in the python path, and also import the working packing (under the speed branch)  with a separate name?

(actual improvement)
image

JasperVanDenBosch:
> Is there a way to load both the package in the python path, and also import the working packing (under the speed branch) with a separate name?

You could make two separate virtualenv and give them respective names, e.g. `env-rsatoolbox-0.3` and `env-rsatoolbox-speedup` to make it easier to switch, but loading them both in the same runtime would involve renaming the package names I think..

smazurchuk:
Hi @HeikoSchuett . So I started using this branch and doing a searchlight analysis. I've noticed this to be the case before too, but it seems that after about 8-10 cores, adding any more cores doesn't increase the speed. Intuitively, this probably amounts to an I/O limitation. If you log into a node running the job and run the `htop` command, you can see that even though CPU usage looks like it's 100% across the cores, the color is red, indicating a potential I/O limitation

image

There is only run process that is all green, so I think that somewhere there is still some python command (like a list comprehension) that is really limiting any gains in speed. For reference, if I just use `pdist` with correlation coefficients, I usually see all cores used and all green (indicating there is no underlying problem accessing numpy arrays to keep up with 24 cores)

Any thoughts or interest in maybe discussing this further?

Thanks!

HeikoSchuett:
Hi @smazurchuk,
that does sound interesting, as it does indicate some issue with the parallelization. It seems to be somewhat different from the issue I am fixing here though and this seems to help already. I'll thus first merge this into main and raise the parallelization issue you raise as a new issue.

PR #215, opened 2021-11-24
scatter plots accept new argument for icon_size

JasperVanDenBosch:
Previously this was hardcoded to be `0.1`


HeikoSchuett (approved):
looks good to me

PR #216, opened 2021-11-26
Reduce memory footprint of shrinkage estimates

### LINKED ISSUE #214: Fallback option for Noise Estimate
In a current experiment we have 300 conditions each of which is sampled 6 times (so 1800 bmaps). A reasonable size ROI (default mode network in left-hemisphere) has about 7,000 vertices in the hcp mesh. At present, it seems this is too large of an ROI for the toolbox to handle calculating the crossnobis distance of. Are there any pointers for using datasets of this size?

This can be recreated with the following code:
```python
import numpy as np
import rsatoolbox as rsa

# Load dataset
dset = np.random.rand(1800,7000)

# Generate Word and presentation numbers
words = []; 
for i in range(1800):
    words.append(f'_{int(i/6)}_')

# Calculate Distance
tmp    = dset
tmp_ds = rsa.data.dataset.Dataset(measurements = tmp,
         descriptors = {'subj':1},
         obs_descriptors = {'words':words})
ds_noise = rsa.data.noise.prec_from_measurements(tmp_ds, obs_desc='words', method='diag')
```
gives
```pre
MemoryError: Unable to allocate 657. GiB for an array with shape (1800, 7000, 7000) and data type float64
```

image

HeikoSchuett:
Improved the implementation of the shrinkage estimates, which kept all inner product matrices in memory originally.

To do so I used the standard formulas for variances computations based on E(x) and E(x^2). These implementations turn out to be faster as well. Thus I decided to always use them.

This resolves issue #214 , where the memory issue was reported first.


HeikoSchuett:
Now extended to address #217 as well.

smazurchuk:
Hi @HeikoSchuett 

Thanks for this fix. I think there are two mistakes? I could open it as a pull request or you could fix them (whatever is easier)

1) 'obs' is hard-coded
https://github.com/rsagroup/rsatoolbox/blob/8014a21ca7cc2a950b6a685fb10c145e420ee8a0/rsatoolbox/data/noise.py#L414

should be
`values, inverse = get_unique_inverse(dataset.obs_descriptors[obs_desc])`

2) The covariance matrix that is returned is sometimes 1 dimensional
https://github.com/rsagroup/rsatoolbox/blob/8014a21ca7cc2a950b6a685fb10c145e420ee8a0/rsatoolbox/data/noise.py#L455

should be 
```python
    else:
        if (cov.ndim == 1):
            cov = np.diag(cov)
        prec = np.linalg.inv(cov)
```

Thanks for the update!

PR #218, opened 2021-12-01
Fix nan

smazurchuk:
So, not the prettiest, but I wasn't really sure of a better approach.

I added an option to the `_parse_input_rdms` that is to just return np.NaNs if their are incompatible NaNs. I then manually check if a NaN was returned in `vector1` and then if it was, just return NaN.

I also added this arguement into the functions that call `compare`, but I didn't update all documentation since it is an optional arguement, and truthfully, most people probably shouldn't use it unless they are doing searchlight analysis (where I did add it to the documentation)

### DIFF
--- a/rsatoolbox/inference/evaluate.py
+++ b/rsatoolbox/inference/evaluate.py
@@ -229,7 +229,7 @@ def dual_bootstrap(models, data, method='cosine', fitter=None,
     return result
 
 
-def eval_fixed(models, data, theta=None, method='cosine'):
+def eval_fixed(models, data, theta=None, method='cosine', returnNaN=False):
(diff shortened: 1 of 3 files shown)


HeikoSchuett:
I have been thinking about this a bit now. I really do not like the solution of removing this check. NaNs in the RDMs happen due to the bootstrapping of conditions for distances between conditions and themselves.
It sounds more sensible to me to remove conditions with NaNs beforehand. That should also be more sensible for a searchlight analysis

smazurchuk:
Hey @HeikoSchuett 

Thanks for the response and the thought. I think that makes sense. It felt a bit haphazard to add this check in

PR #219, opened 2021-12-05
Riemannian distance

mshahbazi1997:
Hello Heiko 
Just finish double checking the code I wrote -

I added riemannian distance function
added a test
added a simple demo (still needs to be more thorough)
make some small change in the plot to make it compatible with riemannian

Any comment from you or the team, I would love to hear and address

Cheers
Mahdiyar


HeikoSchuett (approved):
Ok, so I took the liberty of making a bunch of style changes. This removes the need to add another _all combinations function and makes all the automatic code checks happy.

I think it is now functional and could be merged.

Issue #217, opened 2021-12-01
Unequal Number of Observations per Condition


smazurchuk:
Hi all. It looks to me that this toolbox might have been intended to deal with the case of having unequal numbers of observations per condition as evidenced by the `rsatoolbox.rdm.calc_unbalanced` module. However, it seems to me that there is no way to calculate the mahalanobis distance for unbalanced datasets because there is no function to generate the precision matrix for unbalanced datasets.

This can be demonstrated with the following code 

```python
import numpy as np
import rsatoolbox as rsa

ds = rsa.data.Dataset(
    measurements =  np.random.rand(100,50),
    obs_descriptors = {'cond':np.random.choice(10,size=100)})

prec_matrix = rsa.data.noise.prec_from_measurements(ds,'cond')
```
```
ValueError: all input arrays must have the same shape
```

Would you guys recommend just doing univariate normalization? Perhaps if there are unequal observations per condition but the method is `diag` the toolbox could still do univariate normalization?

Thanks!

HeikoSchuett:
Hi @smazurchuk,
so in principle, the toolbox does already work if you estimate the covariance from residuals or in some other way.

It's a good point that it currently does not allow computing that from unbalanced data. It should be fairly straight forward to implement that, just a little awkward to vectorize. I'll have a look.

smazurchuk:
Thanks for the fix! I'll close the issue once the little fixes are made #221

PR #221, opened 2021-12-14
Fixed two minor mistakes

smazurchuk:
1) Fixed hard coding of 'obs'
2) Added check for cov shape


HeikoSchuett:
Closes #217 
Thanks for removing these bugs!

PR #225, opened 2022-01-13
Fixed Permute RDMs function

### LINKED ISSUE #224: Permute Function doesn't work
Just opening this issue to mention that the permute function in the rdms module doesn't work. This issue can be replicated with the following code

```python
import numpy as np
import rsatoolbox as rsa

ds   = rsa.data.Dataset(measurements = np.random.rand(20,50))
rdm  = rsa.rdm.calc_rdm_correlation(ds)
rdm2 = rsa.rdm.rdms.permute_rdms(rdm)
```
image

smazurchuk:
I think that at some point 'stim' changed to 'index' and this wasn't updated in this function (issue here #224) . Shown here is a working version of this function.

image


HeikoSchuett:
closes #224

Issue #160, opened 2020-11-06
Basing data classes on pandas.DataFrame or xarray.DataArray


jooh:
The project currently contains data classes that are constructed from scratch, with raw numpy arrays to store data and associated dicts for keeping track of meta data. This results in a lot of re-inventing the wheel when it comes to e.g. supporting indexing (see #130) or keeping track of meta data over operations. This kind of stuff is already implemented in a performant and well-tested way in pandas. Is there any interest in adopting this for pyrsa?

Concretely, there are a few ways to do this. We could just use composition, so convert existing data array attributes to dataframes with the descriptor dicts as column labels (we might still want to expose these dicts in the objects but they would not be constructed by quering the DataFrame). Or we could [register the RSA-specific properties and methods as a custom accessor](https://pandas.pydata.org/pandas-docs/stable/development/extending.html). So for instance, the rdm.RDMs object would end up a pandas.DataFrame, but with its current methods and properties in a `df.rdm` attribute. One appealing thing about the latter accessor approach is that our data crunching tools would end up taking DataFrames as input, which makes it a lot easier to combine pyrsa with other data libraries in an analysis pipeline.

There is a learning curve associated with using pandas, especially if you're coming at this from Matlab. But there are huge benefits in terms of offloading complexity to a supporting library, increasing performance, and gaining access to panda's rich feature set.

Would be happy to work a bit on this with e.g. rdm.RDMs as a test case. But it's quite a bit of work so before I do so I wanted to first gauge if there's any interest.

JasperVanDenBosch:
Thanks for thinking about this Johan -  I agree it's good to think about how we structure things.

A few thoughts:

I do like pandas for table-like data. But I'm not sure it's the right fit to replace all our data structures. I've come across a few obscure bugs here and there with it, and have contributed code to the project, and was not particularly impressed with how it's maintained. I also have doubts about the performance aspect - something that we expect to be quite important for pyrsa down the road.

I'm not as familiar with xarray so can't say much about that.

One thing I've been thinking to do for a while; but this is more about the outside than the inside I suppose; is to have wrapper functionality which exposes a pure-array (numpy I was thinking) API. Something like

```python
from pyrsa.ndarray import calc_rdm
rdm_vectors = calc_rdm(data)
```

[...]

Issue #80, opened 2020-07-21
import meadows ma data as DataSet


JasperVanDenBosch:
Should follow after #78 

Instead of `RDMs` this creates `DataSet` objects. Could be used for various meadows tasks

- detection of file type
- ma trials could be an obs_desc
- scope: multiple tasks
- add descriptors from task
- add descriptors from stimuli
- scope: one participant / multiple (MA) tasks (json only)
- scope: multiple participants / multiple (MA) tasks (json only)

JasperVanDenBosch:
Closing this as I now feel this should be in [scythe](https://github.com/meadowsresearch/scythe)

PR #227, opened 2022-02-05
Dataset <--> pandas DataFrame

JasperVanDenBosch:
Proposal for DataFrame conversion of Datasets. This would make it a bit easier to write code for importing table-like data files, such as those stored in event files, CSVs.

### example
```python
ds_in = Dataset(
    measurements=numpy.random.rand(3, 2),
    descriptors=dict(foo='bar'),
    obs_descriptors=dict(abc=['a', 'b', 'c']),
    channel_descriptors=dict(name=['x', 'y'])
)
df = ds_in.to_DataFrame()
ds_out = Dataset.from_DataFrame(df)
print(df)
```
```
          x         y         abc  foo
0  0.709961  0.514268           a  bar
1  0.494080  0.100730           b  bar
2  0.922560  0.530450           c  bar
```

### Related 
- #160 
- #80

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -139,7 +139,8 @@ disable=print-statement,
         deprecated-sys-function,
         exception-escape,
         comprehension-escape,
-        no-self-use
+        no-self-use,
+        use-dict-literal
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/.pylintrc
+++ b/.pylintrc
@@ -415,7 +415,9 @@ good-names=i,
            Run,
            by,
            t,
-           _
+           _,
+           df,
+           ds
 
 # Good variable names regexes, separated by a comma. If names match any regex,
 # they will always be accepted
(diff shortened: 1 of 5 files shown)


HeikoSchuett (commented):
In general this looks good to me. 

Two comments: 

1. Can you add an input to to_DataFrame that allows one to choose which descriptor to use for the channel names? I think that would be a good addition
2. Similarly, I think it would be great if the from_DataFrame function accepted an argument to decide which columns should be come descriptors and which ones should not. People being brought up with numeric matrices will often produce descriptors that are numbers and it might also become frustrating to have integer data not accepted or something like this. Floats are data is a good heuristic default, but not perfect.

JasperVanDenBosch:
@HeikoSchuett Thanks for looking at this! I've added the arguments. I've also changed the method names (`to_df()`, `from_df()`) since codefactor didn't like them (and I suppose this way they are more in line with e.g. `to_dict()`), is this good?

HeikoSchuett (approved):

HeikoSchuett:
I understand that it is generally against doing both camel case and underscores in a name. Also, it is now shorter. I'll just merge this in then!

PR #238, opened 2022-04-14
Bugfix: RDM calculation sorting input data

### LINKED ISSUE #237: calc_rdm_crossnobis changes data structure
calc_rdm_crossnobis actually changes the data structure that the user passes to it. 
On line 305 (rsa.rdm.calc.py) 
dataset.sort_by(descriptor)

The data set is sorted (in place) by descriptor. 

This is very unecexpected and undesirable behavior, as the user may rely in downstream processing that the data is sorted in a specific way. I believe it is also very likely unecessary. 

@HeikoSchuett : Are you maintaining this function? 
Can we change it to ensure the passed data structure is not changed?

HeikoSchuett:
As noticed in #237 the cross validated distance computation sorted the input datasets as a side effect. This pull request removes this problem by copying before the sort.

### DIFF
--- a/rsatoolbox/rdm/calc.py
+++ b/rsatoolbox/rdm/calc.py
@@ -298,6 +298,7 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
     if descriptor is None:
         raise ValueError('descriptor must be a string! Crossvalidation' +
                          'requires multiple measurements to be grouped')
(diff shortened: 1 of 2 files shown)


jdiedrichsen:
Ok - should come back to this at some point to find a more elegant solution - but good for now.

jdiedrichsen (approved):

Issue #236, opened 2022-03-30
cov_from_residuals


mshahbazi1997:
Thanks for considering reporting an issue in `rsatoolbox` to us! Below are some tips that make it easier for us to understand what happened.

**Describe the bug**

**To Reproduce**
Ideally you'd include the smallest code sample required to reproduce the error.

**Expected behavior**

**Plots**
If the bug is related to plotting code, please attach the image(s).

**Versions**
Please report:
- the version of *rsatoolbox* you used (`pip freeze | grep rsatoolbox`)
- the version of python you used (`python --version`)

**Additional context**

**stack trace**
This is the error information that python printed. You can add this in the body of this issue, wrapped in triple backticks (` ``` `) or attach it in a text file

mshahbazi1997:
Hi 
I was using the toolbox for multivariate noise normalization 
the function rsd.noise.cov_from_residuals(res,dof,method='full')
uses numpy.einsum for calculating the covariance. This line of code take ~8 seconds(!) for a matrix of size 4000x2000
matrix = np.random.randn(4000,2000)
mult = np.einsum(‘ij, ik-> jk’, matrix, matrix). but if instead consider changing it to simple matrix product (@) it will be much faster (less than 0.2s on my system). also there is an option in np.einsum  (optimize) if we set it to true it become faster but I am not sure if @ would be faster or einsum with optimize=true. 
hope you find this useful.
cheers

JasperVanDenBosch:
Thanks for reporting this. I can confirm the numbers you reported and with `optimize=True` it becomes about 170ms. It uses a bit more memory though. Also there is some overhead so it may slow down in some use cases but it looks like most of our uses don't fall in that category.
I'd say it looks safe for us to implement this when someone gets around to this. Do you feel like making a PR @mshahbazi1997 ?

mshahbazi1997:
yes - I'd be happy to make a PR.

mshahbazi1997:
but before doing that - I think I noticed another potential issue in the same .py file.
function sample_covariance_3d gives me an error.
mat = np.random.randn(5,60,8)
xt_x = np.einsum('ij, ik-> ijk', mat, mat)
-- I don't if this is a version dependent error or not but when I changed the einsum to 
xt_x = np.einsum('ij..., ik...->...jk', mat, mat)
it works properly.

mshahbazi1997:
@ilogue ?

HeikoSchuett:
[...]

PR #240, opened 2022-04-20
remove deprecated func. make cov calc faster

mshahbazi1997:
Hey - In this pull request I removed a function sample_covariance_3d that wasn't used anywhere in the toolbox and it was depricated. Also I made the rsd.noise.cov_from_residuals(res,dof,method='full') way faster by adding the optimize=True option into the einsum


HeikoSchuett:
closes #236

PR #241, opened 2022-04-25
Structure change in documentation - 

jdiedrichsen:
As discussed with Niko and Heiko - here a restructuring of the 
- Model Specification 
- Model Fitting 
- Model Evaluation & Inference 

Latter part right now is mostly on evaluation - rather than on inference itself. 
The main part is to disucss different forms of evaluation clear. 

@HeikoSchuett : Can you provide a leave-one-out CV example across subjects? 
 
We need to choose whether we need a seperate inference section. Certainly, the model family documentation would best fit in a separate inference section - and the use of bootstrap estimates of variance would be good. We can also explain i directly in the corresponding evaluation section. 

@HeikoSchuett : Can you fill the gaps indicated by the CAPITALS?


HeikoSchuett:
@jdiedrichsen I just fixed the capitalized gaps and in particular fixed the dimensionality bug of fit_regress_nn. It should now all be functional. 

For the cross-validation example you can use:
```
train_set, test_set, ceil_set = rsatoolbox.inference.sets_leave_one_out_rdm(rdms_data)
results_cv = rsatoolbox.inference.crossval(models_flex, rdms_data, train_set, test_set, ceil_set=ceil_set, method='corr')
```
I am not sure where you wanted to put that.

jdiedrichsen:
I included this example in 
inference.rst/Types of Inference / crossvalidation across participants 

The only thing that remains is a bit more information about the bootstrap - I would build it up slowly. 
Fixed evaluated
Crossvalidation across subject 
Bootstrap across subject...
...

@nkriegeskorte : can you review and approve, so we can merge?

PR #246, opened 2022-07-11
deleted faulty indexing from batch_to_vectors

LukasNeugebauer:
There was a bug in ModelFixed that was caused by indexing the function to avoid multiple outputs but still unpacking the output. Should be fixed by deleting the indexing


HeikoSchuett:
Dear @LukasNeugebauer,

thank you for picking up on this!
I added a test to check that this does not get broken again later & will merge this now

best, Heiko

LukasNeugebauer:
Sure, no problem!

PR #251, opened 2022-08-23
Cython port of core similarity computation

HeikoSchuett:
To speed up the computation of dissimilarities, especially for the unbalanced cases I work on porting them into cython such that they are compiled into C code, which allows much faster running of the necessary loops.


rsatoolbox/cutil/similarity.pyx
```diff
@@ -0,0 +1,175 @@
...
+@cython.boundscheck(False)
+cdef double mahalanobis(double [:] vec_i, double [:] vec_j, int n_dim,
+                        double [:, :] noise):
+    cdef:
+        double *vec2
+        int zero = 0
+        int one = 1
+        double onef = 1.0
+        double zerof = 0.0
+        char trans = 'n'
```
JasperVanDenBosch: For this line, `char trans = 'n'`, I'm getting 

> Unicode literals do not support coercion to C types other than Py_UNICODE/Py_UCS4 (for characters) or Py_UNICODE* (for strings).

Did you see the same?

JasperVanDenBosch: Okay managed to make this work by defining it as a byte string

JasperVanDenBosch:
distribution:

- [x] source
- [x] linux
- [x] mac
- [x] windows
- [ ] conda

JasperVanDenBosch:
What do you say we name it something more interesting than `cutil`? How about `engine`? `util` always ends up being the thing where everything ends up and you have no idea what's in there from the name.

HeikoSchuett:
I don't mind changing the name. I would like to keep the information in there that this part is cython / c / compiled code. Maybe something like c_backend?

HeikoSchuett:
@ilogue Now that it finally passes all the automatic tests: Could you have a look whether you think the packaging methods are ok this way?

JasperVanDenBosch:
remaining todos jasper

- [x] appropriately name artifacts
- [x] bump version
- [x] test_c.CalcOne failure test AssertionError: 126 != 105.0 within 7 places (21.0 difference)  C unequal to python for euclidean weight)
- [x] MDS weighted test
- [x] Test_rdm_plot.test_show_rdm_icon_string_no_error
- [x] TestConsistency.test_two_rdms_nn
- [x] test exercises all
- [x] gateway job
- [x] disable profiling
- [x] 3.11
- [x] nan_index
- [ ] list manylinux, other arches
- [ ] cleanup

HeikoSchuett:
OK, I finally succeeded in stabilizing the model fitters such that they do not fail in any significant number anymore. This unfortunately required running multiple starts for the optimization type fitters. 
I will now reduce the number of consistency runs back down to 10 to reduce the test time.

JasperVanDenBosch:
> OK, I finally succeeded in stabilizing the model fitters such that they do not fail in any significant number anymore. This unfortunately required running multiple starts for the optimization type fitters.
> I will now reduce the number of consistency runs back down to 10 to reduce the test time.

Great! Shall we merge now?

Issue #248, opened 2022-07-20
Runtime warnings and errors when using searchlight


jones-michael-s:
Hello,

I'm currently getting runtime warnings and errors when using searchlight (specifically from evaluate_models_searchlight).

**Background:**

• OS-X 10.14.3,  Python 3.7.3,  rsatoolbox 0.0.4, numpy 1.21.6, scipy 1.7.3, pandas 1.3.5, nibabel 4.0.1

• I've been successful in getting basic RSA toolbox functionality to work (i.e., defining a data RDM and a model RDM and fitting the model to the data using  rsatoolbox.inference.eval_fixed). I don't get any warnings or errors and the results look sensible. It's only when attempting to use searchlight that I run into problems.

• As far as I have explored, the problem isn't fixed by options, parameter settings, data size, etc.

**1) runtime warning:**

I  get the following warning when i) running the searchlight example in the documentation or ii) running searchlight on toy data (generated using numpy random):

```
$ ./searchlight_demo.py
Finding searchlights...: 100%|███████████| 61044/61044 [00:14<00:00, 4308.06it/s]
Found 56954 searchlights
Calculating RDMs...: 100%|████████████| 100/100 [05:46<00:00,  3.47s/it]
Evaluating models for each searchlight: 
/usr/local/lib/python3.7/site-packages/rsatoolbox/inference/evaluate.py:255: RuntimeWarning: Degrees of freedom <= 0 for slice
  variances = np.cov(evaluations[0], ddof=1) \
/usr/local/lib/python3.7/site-packages/numpy/lib/function_base.py:2542: RuntimeWarning: divide by zero encountered in true_divide
  c *= np.true_divide(1, fact)
/usr/local/lib/python3.7/site-packages/numpy/lib/function_base.py:2542: RuntimeWarning: invalid value encountered in multiply
  c *= np.true_divide(1, fact)

[ warning repeats two additional times ]

Evaluating models for each searchlight: 100%|██████████| 56954/56954 [01:50<00:00, 517.41it/s]
$
```

The analysis completes, but the divide by zero warning is still concerning.

**2) runtime error**

When trying to run searchlight on real (i.e., our) data I get a different warning and then an error. What's different here is this uses data saved from Matlab and then input using io.matlab.loadmat (code to follow):

```
$ ./searchlight_realdatatest.py
Finding searchlights...: 100%|███████████████████| 254906/254906 [00:22<00:00, 11444.65it/s]
Found 218370 searchlights
[...]

Issue #213, opened 2021-11-22
parallel searchlight does not use more than 8-10 cores efficiently


HeikoSchuett:
Hi @HeikoSchuett . So I started using this branch and doing a searchlight analysis. I've noticed this to be the case before too, but it seems that after about 8-10 cores, adding any more cores doesn't increase the speed. Intuitively, this probably amounts to an I/O limitation. If you log into a node running the job and run the `htop` command, you can see that even though CPU usage looks like it's 100% across the cores, the color is red, indicating a potential I/O limitation

![image](https://user-images.githubusercontent.com/29523859/142681521-412fd008-38b5-4001-ad33-63cb07beda3e.png)

There is only run process that is all green, so I think that somewhere there is still some python command (like a list comprehension) that is really limiting any gains in speed. For reference, if I just use `pdist` with correlation coefficients, I usually see all cores used and all green (indicating there is no underlying problem accessing numpy arrays to keep up with 24 cores)

Any thoughts or interest in maybe discussing this further?

Thanks!

_Originally posted by @smazurchuk in https://github.com/rsagroup/rsatoolbox/issues/212#issuecomment-974365044_

smazurchuk:
I think it might be okay to close this thread because any optimizations to fix this would probably require significant re-writing. For anyone who comes across this thread in the future, I used the [numba](https://numba.pydata.org/) toolbox to make a version of the `crossnobis` function that compiles in no-python mode. Even with a completely jitted function, this speed limitation exists. The kernel threads can be moved into the userspace by avoiding the use of `joblib` and just using a list comprehension. The speed ends up being about the same.

![image](https://user-images.githubusercontent.com/29523859/142946109-bc1b9420-d032-4921-a327-d31b47e1928c.png)

As an aside, doing this calculation with only numpy arrays and no python code, the speed gain is only about **4x** from the current main branch. I think this gives a good upper limit for the fastest the calculation could be on CPU's. Linked here is a notebook demonstrating this. For reference, the fast version without the `@njit` decorator takes 15 seconds

https://github.com/smazurchuk/general/blob/master/fastnobis.ipynb

HeikoSchuett:
I'll leave this open for now. I am glad that it does not look like extreme losses of speed, but the code for the searchlight really should be embarrassingly parallel, i.e. if implemented optimally this should scale perfectly with more cpus.

[...]

PR #253, opened 2022-08-25
searchlight revisited

JasperVanDenBosch:
Rethink of the searchlight interface, architecture as well as introducing a surface-based version.

## datasets

- sitek (private)
- jones-michael-s (private)
- cichy16 (public)
- jsong (private)

## issues

- #248 
- #292
- #272
- #213

## thoughts and notes

Performance is important for searchlight analysis. The current interface does not readily allow or isn't optimised for an approach where each chunk of data goes through multiple steps in a pipeline; but this would enable parallelisation on an HPC infrastructure. An approach with in-process parallelization using e.g. job lib brings certain complications as well.


iancharest:
i had to create an intermediate calc_rdm_batch function to deal with the fact that calc_rdm involves recursion and with parallel, this lead to maximum recursive depth errors.

PR #260, opened 2022-09-15
Fix docstring typo in rdm_plot.py show_rdm

charlesbmi:
`show_colorbar` expects `'panel'` or `'figure'`: https://github.com/rsagroup/rsatoolbox/blob/de72c74/rsatoolbox/vis/rdm_plot.py#L94


HeikoSchuett:
thanks for picking this up

PR #269, opened 2022-10-13
Inference output

### LINKED ISSUE #264: Extracting and visualizing statistical inference
Thanks for the great toolbox, I've quite enjoyed using it. I have a suggestion that I think might improve the usability of the statistical inference procedures.

**Is your feature request related to a problem? Please describe.**
Once an object of class `rsatoolbox.inference.result.Result` is created, it is difficult to extract (e.g.) a table of statistics from running inferential tests. It is very straightforward to visualize these statistics using `rsatoolbox.vis.plot_model_comparison`, but I often find myself wanting direct access to the numbers. This is helpful for (e.g.) reporting the results of specific statistical tests in the text of a manuscript, or for creating custom plots highlighting key results.

Related to that second point, `plot_model_comparison` currently does not return the created figure. Therefore, while it's easy to create an inline plot of model comparisons in a jupyter notebook, it's difficult to save that plot to disk (or to customize it).

**Describe the solution you'd like**
As currently instantiated, `plot_model_comparison` is really doing the work of three functions at once: getting p-values from the relevant statistical tests (one-sample and pairwise tests), computing variance around the expectations (e.g., SEM, CI), and then plotting all of the results.

As an alternative, these operations could be refactored into two separate functions. The first new function might return a human-readable dictionary (or table) of all statistical results. To illustrate what this might look like:

```python
{
    "noise_ceil_lower": value,
    "noise_ceil_upper": value,
    "pvalue_model0_vs_0": p-value testing model0 against 0,
    "pvalue_model0_vs_model1": p-value from pairwise test of model0 against model1,
    etc...
}
```

Once this dictionary/table of results is created, it could then be passed into a new plotting function, which basically is just the plotting elements of `plot_model_comparison` shunted off into a new function. To preserve backwards compatibility and user-friendliness for those who don't want/need access to the statistics, `plot_model_comparison` might then be a thin wrapper around the two refactored functions.

**Volunteer**
- would you be willing to write code for the library with us?
- if so, what do you need from us to get started?

In principle, I would be willing to write code to address this issue. However, I should disclaim that I'm very much a software user, not a developer. I have relatively little familiarity with software development (e.g., modifying "published" libraries, writing tests, etc), and so I would need some support from developers to learn how to be an effective contributor.

HeikoSchuett:
This pull request adds functions to extract the test results and a print functionality to the results objects.

This should resolve the issues described in #264

### DIFF
--- a/rsatoolbox/inference/result.py
+++ b/rsatoolbox/inference/result.py
@@ -94,7 +94,7 @@ class Result:
         """
         summary = f'Results for running {self.cv_method} evaluation for {self.method} '
         summary += f'on {self.n_model} models:\n\n'
-        name_length = max([max([len(m.name) for m in self.models]) + 1, 6])
+        name_length = max([max(len(m.name) for m in self.models) + 1, 6])
         means = self.get_means()
         sems = self.get_sem()
         p_zero = self.test_zero(test_type=test_type)
@@ -183,12 +183,15 @@ class Result:
         return p_pairwise, p_zero, p_noise
 
     def test_pairwise(self, test_type='t-test'):
+        """returns the pairwise test p-values """
         return pair_tests(self.evaluations, test_type, self.diff_var, self.dof)
 
     def test_zero(self, test_type='t-test'):
+        """returns the p-values for the tests against 0 """
         return zero_tests(self.evaluations, test_type, self.model_var, self.dof)
 
     def test_noise(self, test_type='t-test'):
+        """returns the p-values for the tests against the noise ceiling"""
         return nc_tests(self.evaluations, self.noise_ceiling,
                         test_type, self.noise_ceil_var, self.dof)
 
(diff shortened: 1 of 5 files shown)


rsatoolbox/inference/result.py
```diff
@@ -67,6 +70,63 @@ def __init__(self, models, evaluations, method, cv_method, noise_ceiling,
...
+            test_type(String):
+                What kind of tests to run.
+                See rsatoolbox.util.inference_util.all_tests for options
+        """
+        summary = f'Results for running {self.cv_method} evaluation for {self.method} '
+        summary += f'on {self.n_model} models:\n\n'
+        name_length = max([max([len(m.name) for m in self.models]) + 1, 6])
+        means = self.get_means()
+        sems = self.get_sem()
+        p_zero = self.test_zero(test_type='t-test')
```
psychNerdJae: I believe this should be something like `self.test_zero(test_type=test_type)`, in cases where the user asks for something other than a t-test?

HeikoSchuett: yes, it should be!

PR #285, opened 2022-11-29
Change default colormap

### LINKED ISSUE #232: Colormap unification
We talked last time about a unification fo the colormaps used in the toolbox. We certainly want to keep the option for users to choose their own colormaps of course, but by default we should have a reasonably consistent setting

JasperVanDenBosch:
resolves #232 

- [x] script to showcase colormaps
- [x] send around poll
- [x] evaluate
- [x] remove scripts
- [x] new default
- [x] move "classic" map


JasperVanDenBosch:
-- sample removed; extensive one below --

nkriegeskorte:
several of these (all?) are designed to be perceptually uniform. that is a feature we should go for.

however, there are some additional considerations that might motivate other colormaps:

(1) edward tufte argues (compellingly, i think) that more ink on a white page should be associated with larger values.
this suggests that we should use a default colormap that goes from light to dark.

(2) in our figures,
 we often want to use color to encode conditions, condition sets, models, or subsets of significant results.
this suggests using an achromatic or minimally chromatic scale like
- greys
- gray
- gist_gray
- gist_yarg
- bone

perhaps these are sufficiently close to perceptually uniform, or we could adjust them for perceptual uniformity?

niko

2022-11-29 16_32_48-Choosing Colormaps in Matplotlib — Matplotlib 3 6 2 documentation

JasperVanDenBosch:
rdm_colormaps_dark2light
rdm_colormaps_light2dark

PR #293, opened 2023-01-21
small bug fix for list descriptors in RDM calculations

HeikoSchuett:
List descriptors in the original dataset were breaking _build_rdms
This can be fixed with a simple np.asarray as I implement here.


JasperVanDenBosch:
Surprised as I thought I had taken this into account for `_build_rdms()`. I'll add a regression test when I get a chance. But the pipeline runs so that's a good sign.

HeikoSchuett:
it surprised me, too, because I didn't run into any errors for quite a while, but
orig_obs_desc_vals == v
is just False if orig_obs_desc_vals is a list.

JasperVanDenBosch:
I can't get it to fail. Could you use something from the situation where it happened to improve the test? I'd just sleep easier knowing we have a test that catches it next time.

PR #302, opened 2023-02-28
Conda packaging

### LINKED ISSUE #258: Enable install through conda
**Is your feature request related to a problem? Please describe.**
At the moment, the package cannot be installed through conda/anaconda because it is not available there.
Also, `petname==2.2` is not available there, but required.

**Describe the solution you'd like**
Include petname=2.2 and make a conda package to enable conda installs.

**Describe alternatives you've considered**
pip install and conda install mixed yields environments that neither method can resolve properly at some point.

JasperVanDenBosch:
TBD Conda-forge submission

resolves #258 

### todo

- [x] skeleton tests
- [x] `build_locally` passes
- [x] `lint_recipe` passes
- [x] forge PR

### DIFF
--- a/.conda/README.md
+++ b/.conda/README.md
@@ -1,0 +1,5 @@
+
+in run_docker_build.sh line 42:
+```
+DOCKER_RUN_ARGS="-it --dns 8.8.8.8"
+```
--- a/.conda/README.md
+++ b/.conda/README.md
@@ -1,5 +1,0 @@
-
-in run_docker_build.sh line 42:
-```
-DOCKER_RUN_ARGS="-it --dns 8.8.8.8"
-```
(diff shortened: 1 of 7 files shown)


JasperVanDenBosch:
Currently fails as Cython 3 is not on conda yet

Relevant threads:
- https://github.com/cython/cython/issues/4022
- https://github.com/conda-forge/cython-feedstock/pull/75

h-vetinari:
> Currently fails as Cython 3 is not on conda yet

You can actually install the cython 3 beta from conda-forge as follows (basically you need to specify a separate channel, which is also possible in environment files):
```
conda install -c conda-forge/label/cython_dev cython=3
```
You can also see the available builds [here](https://anaconda.org/conda-forge/cython/files)

JasperVanDenBosch:
> You can actually install the cython 3 beta from conda-forge as follows (basically you need to specify a separate channel, which is also possible in environment files):

Thanks! Turns out it also builds well on the latest Cython 2 version so I was able to relax that constraint.

PR #303, opened 2023-03-03
Integer input for unbalanced calc functions

JasperVanDenBosch:
It seems the the Cython code expects floats, and the python wrappers do not reject or handle integers.

### DIFF
--- a/src/rsatoolbox/rdm/calc_unbalanced.py
+++ b/src/rsatoolbox/rdm/calc_unbalanced.py
@@ -116,7 +116,8 @@ def calc_rdm_unbalanced(dataset: SingleOrMultiDataset, method='euclidean',
             weight_idx = 1
         cond_indices_int = cond_indices.astype(int)
         rdm = calc(
-            dataset.measurements, cond_indices_int,
+            ensure_double(dataset.measurements),
+            cond_indices_int,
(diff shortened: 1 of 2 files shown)


JasperVanDenBosch:
@HeikoSchuett the type casting seems to be working as intended. 

I seem to misunderstand something about the weighting and cv descriptors in `calc_one` (see failing test). Shouldn't it return the same dissimilarity as `calc_rdm_unbalanced`?

Also see this line, it doesn't seem to have an effect (the cv descriptors aren't used here, and refer to the "parent" dataset): 

https://github.com/rsagroup/rsatoolbox/blob/b811fb8037f4dced82d0d9afbba23e6892d77aa4/src/rsatoolbox/cengine/similarity.pyx#L166

HeikoSchuett:
Hi @ilogue,
this actually looks correct to me. It just divides by the number of channels already. The thing it computes is the similarity kernel, i.e. 0*0 + 2*2 in your case and then divides by the number of channels -> =2

The line with the cv_descriptors is in there, because I wrote this function originally to become a part of the rdm calculation. It still makes sense to have this to remove observations from the same session from the similarity computation. However, calc_one is now not actually used in the rdm calculation and I kept it just in case it is helpful sometimes. I don't think it comes up in many workflows.

PR #315, opened 2023-03-30
Visualize sort

### LINKED ISSUE #314: Model comparison plot can not take `sort=True`
When using `plot_model_comparison`, we get the following error if using more than 1 model with `sort!=False`. The following trace is seen. Digging a bit more, the error seems to be caused by `model_var` being a np array. An easy fix would be to use the longer check of `if model_var is not None`.

```
in plot_model_comparison(result, sort, colors, alpha, test_pair_comparisons, multiple_pair_testing, test_above_0, test_below_noise_ceil, error_bars, test_type)
    256 perf = perf[idx]
    257 evaluations = evaluations[:, idx]
--> 258 if model_var:
    259     model_var = model_var[idx]
    260 if noise_ceil_var:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Aceticia:
Fixes #314

### DIFF
--- a/src/rsatoolbox/vis/model_plot.py
+++ b/src/rsatoolbox/vis/model_plot.py
@@ -336,7 +336,8 @@ def plot_model_comparison(result, sort=False, colors=None,
     else:
         ax.bar(np.arange(evaluations.shape[1]), perf, color=colors)
     if error_bars:
-        limits = get_errorbars(model_var, evaluations, dof, error_bars, test_type)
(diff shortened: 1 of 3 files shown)


HeikoSchuett (approved):
Thank you for catching this bug, too. It if works for you,  I am happy to merge it in.

JasperVanDenBosch:
If that's ok I'd like to add a test to this. I'll get around to it later today and will add a review after

HeikoSchuett (approved):

PR #316, opened 2023-04-04
allow python 3.11

HeikoSchuett:
When setting up my new computer I noticed that python 3.11 is now the default installed version, which rsatoolbox currently does not allow. Here I'll setup things to make it accept and test python 3.11


JasperVanDenBosch (approved):
Looks like all our dependencies now have 3.11 compatible versions! This looks good to me!

PR #319, opened 2023-04-18
fixing seed for test runs

HeikoSchuett:
As we are running into test problems from time to time due to rare failures of the optimizations to converge, I here fix the seed for the random tests. This will give us largely consistent test results throughout. For consistency, I did this everywhere.


JasperVanDenBosch (approved):
Looks good!

PR #331, opened 2023-06-28
added possibility to run calc_unbalanced for rdm movies

HeikoSchuett:
It came up in support mail recently, that there is no way to run the unbalanced calculations to creat RDM movies. This pull request adds a flag to calc_rdm_movie to allow this and then just runs calc_rdm_unbalanced instead of calc_rdm.

This requiered a little bit of restructuring of helper functions to prevent circular imports between calc_rdm and calc_rdm unbalanced.


JasperVanDenBosch (approved):
LGTM

Issue #348, opened 2023-09-21
Spearmanr comparison different from documentation


Aceticia:
The implementation at https://github.com/rsagroup/rsatoolbox/blob/main/src/rsatoolbox/rdm/compare.py#L185 seems to be different from that in the documentation: https://rsatoolbox.readthedocs.io/en/stable/comparing.html#spearman-s-rho. Is this intended?

HeikoSchuett:
The two are actually equivalent. 
The function in the toolbox calculates 12 * (x-E(x))^t(y-E(y)) / (n^3-n), because E(x) = E(y) = (n+1)/2 this can be transformed into the formula in the documentation and paper:
12 * (x-E(x))^t(y-E(y)) / (n^3-n)
= 12 * (x^ty - 2n(n+1)^2/4 + n(n+1)^2/4 )  / (n^3-n) 
=  12 * (x^ty )  / (n^3-n) -  3*n(n+1)^2  / (n^3-n)
= 12 * (x^ty )  / (n^3-n) -  3*(n+1)  / (n-1)
which is the formula in the paper

It is not an intentional choice, but you get the same results.

Aceticia:
My confusion is that the doc has no rank operation, but the code does. Is it that the doc's 'x, y' refers to rank, not raw input?

HeikoSchuett:
it should be rank transformed, yes. Sorry that this was unclear. I just checked and in the paper it says that at least. I will make a change in the docs here to say that too.

PR #349, opened 2023-09-25
Update comparing.rst

HeikoSchuett:
update to include the information that we need to rank transform the RDMs for the spearman rank correlation  of course


JasperVanDenBosch (approved):

JasperVanDenBosch:
resolves #348

PR #350, opened 2023-09-26
Set `rdm.dissimilarity_measure` in `calc_rdm_movie`.

### LINKED ISSUE #344: calc_rdm_movie's returned RDMs doesn't have its .dissimilarity_measure set
**Describe the bug**
`calc_rdm_movie` accepts a `method` parameter, which it uses for all the matrices in the `RDMs` object it returns, guaranteeing that they all use the same method. As such, it could (should) set the returned `RDMs.dissimilarity_measure` to be the `method`.

**To Reproduce**
(`data` as in the [Temporal RSA demo](https://rsatoolbox.readthedocs.io/en/stable/demo_temporal.html))
```python
rdms_data = calc_rdm_movie(data, method="correlation", descriptor="conditions")
print(rdms_data.dissimilarity_measure)  # None
```

**Expected behavior**
Prints `"correlation"`.

**Versions**
- the version of *rsatoolbox* you used
    - 0.1.3
- the version of python you used
    - 3.11.5

caiw:
Fixes #344.

### DIFF
--- a/src/rsatoolbox/rdm/calc.py
+++ b/src/rsatoolbox/rdm/calc.py
@@ -164,6 +164,7 @@ def calc_rdm_movie(
 
         rdm = concat(rdms)
         rdm.rdm_descriptors[time_descriptor] = time
+    rdm.dissimilarity_measure = method
     return rdm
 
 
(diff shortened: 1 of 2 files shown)


JasperVanDenBosch (approved):
Thanks!

PR #353, opened 2023-10-10
`show_rdms(..., nanmask=None)` doesn't suppress any elements

### LINKED ISSUE #352: Easy way to suppress `show_rdms(..., nanmask=)`
My personal preference is not to erase the diagonals from my RDM figures, especially when they are shown on a colour scale in which white is 1 and black is 0 (where the diagonal "should" be black, but would appear white when nanmasked).

However the default behaviour of `show_rdms`'s `nanmask` argument is to erase the diagonal.  The only way to suppress this is something like
```python
show_rdms(my_rdms, ..., nanmask=numpy.zeros_like(my_rdms[0].get_matrices().squeeze(), dtype=bool))
```
which is annoying to do each time.

It is arguably violating the rule of least-surprise to have `nanmask=None` apply a diagonal mask, rather than no mask.

It would be great to have either:
1. The default behaviour be no masking.
2. Some special strings which set up the masking for common cases, e.g. `nanmask="diagonal"`, `nanmask="none"` (or, better, change `nanmask=None` to be no masking).

caiw:
Addresses #352.

- Changes behaviour of `show_rdms` so that argument `nanmask=None` doesn't suppress display of any elements. (Previously it suppressed the diagonal).
- Adds new `nanmask="diagonal"` option to suppress the diagonal.
- Does not change the default behaviour when not specifying `nanmask`.

### DIFF
--- a/src/rsatoolbox/vis/rdm_plot.py
+++ b/src/rsatoolbox/vis/rdm_plot.py
@@ -34,7 +34,7 @@ def show_rdm(
     gridlines: npt.ArrayLike = None,
     num_pattern_groups: int = None,
     figsize: Tuple[float, float] = None,
-    nanmask: npt.ArrayLike = None,
+    nanmask: npt.ArrayLike | str | None = "diagonal",
     style: Union[str, pathlib.Path] = RDM_STYLE,
     vmin: float = None,
     vmax: float = None,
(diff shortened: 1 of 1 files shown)


HeikoSchuett (approved):
While I do not have strong opinions on the default behaviour here, I think the argument makes sense and the implementation looks fine to me.

JasperVanDenBosch (approved):
Thanks Cai!

PR #360, opened 2023-11-09
Mean remove RDMs

HeikoSchuett:
This pull request implements the distances that first remove the mean for each pattern and then compute the respective distances. I made this work for all the variants of euclidean distances by tweaking the input parsing function.


JasperVanDenBosch (approved):
👍

PR #372, opened 2023-11-21
Fix a bug which could mutate a dataset when computing a covariance matrix

### LINKED ISSUE #371: `cov_from_unbalanced(dataset, ...)` mutates `dataset` in place
Lines 392 to 395 [read](https://github.com/rsagroup/rsatoolbox/blob/8645f347128bdbe633a95a2259594323b9e31125/src/rsatoolbox/data/noise.py#L392-L395):
```python
matrix = dataset.measurements
means, values, _ = average_dataset_by(dataset, obs_desc)
values, inverse = get_unique_inverse(dataset.obs_descriptors[obs_desc])
matrix -= means[inverse]
```

In the first line, `matrix` stores a reference to `dataset.measurements`; on the last line that reference is modified in place (because of how numpy ndarrays handle `__isub__`).

It would be fixed by a simple `matrix = dataset.measurements.copy()` on the first line.

caiw:
Fixes #371


JasperVanDenBosch (approved):
Thanks @caiw !

PR #375, opened 2023-11-27
`RDMs.sort_by(reindex)` argument

### LINKED ISSUE #358: `RDMs.sort_by()` take a `reindex` boolean argument to replace the `"index"` pattern descriptor
I just became rather confused because `RDMs.sort_by()` followed by `RDMs.subset_patterns(by="index", ...)` didn't have the expected effect.  In other words, sorting the conditions _also_ sorts `.pattern_descriptors["index"]`, so then `.subset_patterns(by="index", ...)` doesn't subset the patterns by their _current_ indices, but by their indices _prior to sorting_.

I'm sure there are many cases where it makes sense to sort the `"index"` pattern descriptor too, but in my case it'd be useful to automatically keep the patterns with an incremental index.  I propose `RDMs.sort_by()` takes a new boolean argument `reindex` (default `False`), which when `True` resets the `"index"` pattern descriptor to 0, 1, ..., n_cond-1.

Is there a reason or current usage of the `"index"` pattern descriptor which would make this problematic?

caiw:
Adds a new `reindex: bool` argument to `RDMs.sort_by()`, default `True`, which resets the index after sorting.

@JasperVanDenBosch [suggested](https://github.com/rsagroup/rsatoolbox/issues/358#issuecomment-1801824285) that it also throw a depricationwarning, but I'm not sure what the policy is on this.

Fixes #358.

## Some issues

If the RDMs has a `pattern_descriptor` named "reindex", this will upset the function. I'm not sure what to do about this, other than change the signature of `sort_by()`.

### DIFF
--- a/src/rsatoolbox/rdm/rdms.py
+++ b/src/rsatoolbox/rdm/rdms.py
@@ -422,8 +422,12 @@ class RDMs:
         for dname, descriptors in self.pattern_descriptors.items():
             self.pattern_descriptors[dname] = [descriptors[idx] for idx in new_order]
 
-    def sort_by(self, **kwargs):
+    def sort_by(self, reindex: bool = True, **kwargs):
         """Reorder the patterns by sorting a descriptor
+
+        Args:
+            reindex (bool): whether to reset the 'index' descriptor 
+                following sorting
 
         Pass keyword arguments that correspond to descriptors,
         with value indicating the sort type. Supported methods:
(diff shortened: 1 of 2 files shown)


JasperVanDenBosch:
Let's just assume `reindex` is an unlikely descriptor name?

And I guess the warning would be a bit tricky to set up here.

JasperVanDenBosch (approved):

PR #376, opened 2023-11-27
RDMs.sort_by is stable

### LINKED ISSUE #374: 'RDMs.sort_by' sorting is not stable
Not sure if this is a bug-report, as I'd argue the current behaviour is unexpected, or if it's just an enhancement request.

The sorting algorithm used in `RDMs.sort_by` is not [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability), i.e. equal elements aren't guaranteed to have their order preserved.  By contrast, Python's `sorted()` is stable.  I'm not sure about numpy, but presumably it's not because I think that's what `sort_by` is using.

## Why one would care about stability

Suppose we have a `pattern_descriptor` called "object" which stores a string name of image photos shown to participants, and a second `pattern_descriptor` called "category" which stores a string name of one of 5 object categories relating the the image.  So the "object" `pattern_descriptor` might store "alligator", "giraffe", "daisy", "rose", etc. for 4 conditions; and the "category" `pattern_descriptor` might store "animal", "animal", "flower", "flower" for the same 4 conditions.

Suppose I want to order the conditions in the RDM so that all animals appear at the top, then all flowers, etc.  And within each category block, the items to be ordered alphabetically.  So for the above 4 conditions this should be:
- [animal] alligator
- [animal] giraffe
- [flower] daisy
- [flower] rose
etc.

My expected way to do this is a 2-level sort like:

```python
rdms.sort_by(**{"object": "alpha"})  # Now all the objects are in alpha order by the categories are jumbled
rdms.sort_by(**{"category": "alpha"})  # Now all the categories are in alpha order, and where there are ties (same category) the previous sort is preserved
```

However this doesn't (always) work.
## How to workaround it

Currently I'm using the folliwing few helper functions:

```python
def argsort(seq):
    """Pure python argsort"""
    # Thanks https://stackoverflow.com/a/57028877/2883198
    return sorted(range(len(seq)), key=seq.__getitem__)

def reset_pattern_index(rdms: RDMs):
    """Resets rdms.pattern_descriptors["index"] in place."""
    rdms.pattern_descriptors["index"] = list(range(rdms.n_cond))

def sort_bys(rdms: RDMs, pattern_descriptors: tuple[str, ...], reindex: bool = True):
    """Sort by each of the pattern_descriptors, alpha order, in order (i.e. ties in the first resolved by the second, etc.)"""
    sorting_data = list(zip(*(rdms.pattern_descriptors[d] for d in pattern_descriptors)))
    idxs = argsort(sorting_data)
    rdms.sort_by(**{"index": idxs})
    if reindex:
        reset_pattern_index(rdms)

# Use like
sort_bys(rdms, pattern_descriptors=("category", "object"))
```
 
(As you can see, part of the above is also my workaround for #358)

I'm dropping the above here because I wonder if these multi-level sorts could also be incorporated into `sort_by`, which currently has (to my eye) a slightly strange interface.

caiw:
Makes `RDMs.sort_by` sort stable by passing a stable specification to numpy.
Adds a basic test for stability.

Fixes #374.

### DIFF
--- a/src/rsatoolbox/rdm/rdms.py
+++ b/src/rsatoolbox/rdm/rdms.py
@@ -454,7 +454,7 @@ class RDMs:
         for dname, method in kwargs.items():
             if method == 'alpha':
                 descriptor = self.pattern_descriptors[dname]
-                self.reorder(np.argsort(descriptor))
+                self.reorder(np.argsort(descriptor, kind='stable'))
             elif isinstance(method, (list, np.ndarray)):
                 # in this case, `method` is the desired descriptor order
                 new_order = method
--- a/tests/test_rdm.py
+++ b/tests/test_rdm.py
@@ -460,6 +460,35 @@ class TestRDM(unittest.TestCase):
             )
         )
 
+    def test_sort_stable(self):
+        from rsatoolbox.rdm import RDMs
+        rdm = np.array([
+            [0., 1., 2., 3., 4., 5., 6., 7., 8., 9.],
+            [1., 0., 1., 2., 3., 4., 5., 6., 7., 8.],
+            [2., 1., 0., 1., 2., 3., 4., 5., 6., 7.],
+            [3., 2., 1., 0., 1., 2., 3., 4., 5., 6.],
+            [4., 3., 2., 1., 0., 1., 2., 3., 4., 5.],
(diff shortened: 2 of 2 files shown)


JasperVanDenBosch (approved):
LGTM

PR #377, opened 2023-11-27
concat sets dissimilarity_measure if it is identical for all rdms in list

### LINKED ISSUE #368: Concat() should set dissimilarity measure if it is identical
Concat currently does not set the `dissimilarity_measure` attribute

caiw:
Checks for all dissimilarity_measures being the same and sets it if so.

Also add test for the above.

Not testing that it's not set when different.

Fixes #368.


JasperVanDenBosch (approved):
LGTM, thanks Cai!

PR #392, opened 2024-06-02
[FEAT] Topological RSA

doerlbh:
Incorporated necessary functions for Topological RSA.

**Reference**: Lin, B., & Kriegeskorte, N. (2023). The Topology and Geometry of Neural Representations. arXiv preprint arXiv:2309.11028.

**Summary of added functions:**
- minmax_transform(rdms)
- geotopological_transform(rdms, l, u)
- geodesic_transform(rdms)

**Summary of changes:**
- rdm/transform.py
- rdm/\__init\__.py
- test_rdm.py
- requirements.txt

Thank you!
Baihan

### DIFF
--- a/tests/test_rdm.py
+++ b/tests/test_rdm.py
@@ -323,7 +323,7 @@ class TestRDM(unittest.TestCase):
 
     def test_geodesic_transform(self):
         from rsatoolbox.rdm import geodesic_transform
-        dis = np.zeros((8, 10))
+        dis = np.random.rand(8, 10)
         mes = "Euclidean"
         des = {'subj': 0}
         pattern_des = {'type': np.array([0, 1, 2, 2, 4])}
--- a/tests/test_rdm.py
+++ b/tests/test_rdm.py
@@ -289,6 +289,54 @@ class TestRDM(unittest.TestCase):
         self.assertEqual(pos_rdm.n_cond, rdms.n_cond)
         assert np.all(pos_rdm.dissimilarities >= 0)
 
+    def test_minmax_transform(self):
+        from rsatoolbox.rdm import minmax_transform
+        dis = np.zeros((8, 10))
+        mes = "Euclidean"
+        des = {'subj': 0}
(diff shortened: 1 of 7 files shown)


tests/test_rdm.py
```diff
@@ -289,6 +289,54 @@ def test_positive_transform(self):
...
+        des = {'subj': 0}
+        pattern_des = {'type': np.array([0, 1, 2, 2, 4])}
+        rdm_des = {'session': np.array([0, 1, 2, 2, 4, 5, 6, 7])}
+        rdms = rsr.RDMs(dissimilarities=dis,
+                        rdm_descriptors=rdm_des,
+                        pattern_descriptors=pattern_des,
+                        dissimilarity_measure=mes,
+                        descriptors=des)
+        mm_rdm = minmax_transform(rdms)
+        self.assertEqual(mm_rdm.n_rdm, rdms.n_rdm)
```
JasperVanDenBosch, line 305: I don't think these tests for the number of rdms and conditions add much here. My suggestions would be to remove them.

JasperVanDenBosch (approved):
Hi Baihan,
Thanks for this. I've made some small additions. Thanks for adding unit tests but I'm not in favor of doing the transforms on all-zero rdms and then testing their dimensions. I'd suggest leaving these out until such a time as someone has time to write a test that tests the actual transformed values.
Other than that (and Heiko's comment on the scope of the transform) I have no objections
Jasper

tests/test_rdm.py
doerlbh: Hi Jasper, thanks for the review! I see that the sqrt_transform and rank_transform has a similar test (which was what I based my test upon). My thinking is that it can still catch some issues, which without it, cannot find -- e.g. that calling the functions will not throw an error. I fully agree that, with more time, actually adding a test element regarding the values itself would be more helpful.

HeikoSchuett (approved):
This seems fine to me, assuming you checked that this is equivalent to the functions used for the tests in the paper.

PR #399, opened 2024-07-02
Updating GitHub actions & types

HeikoSchuett:
Recently GitHub actions started to fail and at least one of them is due to some types going away in numpy 2.0. Here I fix this

### DIFF
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -13,10 +13,10 @@ jobs:
   tests:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - uses: actions/setup-python@v4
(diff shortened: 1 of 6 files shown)


HeikoSchuett:
ok, now the checks are passing again! please have a look @JasperVanDenBosch 

Summary of changes:
- numpy 2 does no longer have np.string_, replaced that
- some windows versions apparently switch in numpy to int128 as a default? not the type that corresponds to long in c at least.
- adjusted the c-code and setup.py to adjust the types consistently to float64 and int64 for interactions with Cython
- updated the Github actions such that the compile errors go away

JasperVanDenBosch (approved):
Looks great!

JasperVanDenBosch:
> * some windows versions apparently switch in numpy to int128 as a default? not the type that corresponds to long in c at least.

I know that for plain python the underlying types adjust size based on the machine, so in a way that makes sense. They try to make sure it works across all architectures including IoT

PR #400, opened 2024-07-04
Bures similarity and distance

HeikoSchuett:
Adding the Bures similarity and distance as introduced by Sarah E. Harvey, Brett W. Larsen, Alex H. Williams
Duality of Bures and Shape Distances with Implications for Comparing Neural Representations


JasperVanDenBosch (approved):
Looks great!

PR #438, opened 2025-05-24
bug fix for non-functional spm_filter()

bogpetre:
spm_filter assigned fdata to Y, modified Y, and then returned fdata, suggesting Y was meant to modify fdata in place, but Y was a copy, not a view of fdata, so fdata was never updated at all and spm_filter simply returned its input unmodified. The fix implements an inplace updating of fdata instead.

### DIFF
--- a/src/rsatoolbox/io/spm.py
+++ b/src/rsatoolbox/io/spm.py
@@ -149,6 +149,6 @@ class SpmGlm:
 
         fdata = data.copy()
         for i in range(self.nruns):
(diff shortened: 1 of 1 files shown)


JasperVanDenBosch (approved):
Good catch. Thanks Bogdan!

JasperVanDenBosch:
This is now included in the latest dev build. Install with either:
- `pip install rsatoolbox==0.2.1.dev49`
- `pip install --pre rsatoolbox`

PR #453, opened 2025-08-03
Update compare.py._parse_input_rdms

### LINKED ISSUE #452: _parse_input_rdms check for matching nan positions fails
Thanks for considering reporting an issue in `rsatoolbox` to us! Below are some tips that make it easier for us to understand what happened.

**Describe the bug**
_parse_input_rdms checks the two input RDMs only for equal number of NaNs, not that their positions match.

**To Reproduce**
import rsatoolbox as rsa
import numpy as np
vec1 = np.array([1, np.nan, 3])
vec2 = np.array([np.nan, 2, 3])
vec3 = np.array([1,2,3])
rsa.rdm.compare(vec1, vec2)
>>> array([[0.96476382]])
rsa.rdm.compare(vec1, vec3)
>>> ValueError: rdm1 and rdm2 have different nan positions

**Expected behavior**
vec1 and vec2 should raise the error as well.

The current behavior removes the nans and condenses the vectors down to their size without the NaNs. It then checks only that the vectors are the same size after condensing, finds they are (in the problem case), and returns them. Given the importance of position alignment for distance measurements, this seems problematic.

I'll submit a PR shortly that corrects the check to enforce identically placed NaNs.

JonathanShor:
Ensure nan positions in each input RDM match.

Addresses #452

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -29,8 +29,8 @@ jobs:
       run: echo "$(ruff check -e --output-format github)"
     - name: Process Ruff output
       run: python devops/style_assistant.py ${{ github.run_id }} ${{ github.job }} ${{ github.event.pull_request.head.sha }}
-    - name: Report status
-      run: gh api ${{ env.DEVOPS_ASST_API_ARGS }}
+    # - name: Report status
+    #   run: gh api ${{ env.DEVOPS_ASST_API_ARGS }}
 
 
   typing:
@@ -75,5 +75,5 @@ jobs:
       run: pyright > pyright_pr.log
     - name: Process Pyright output
       run: python devops/typing_assistant.py ${{ github.run_id }} ${{ github.job }} ${{ github.event.pull_request.head.sha }}
-    - name: Report status
-      run: gh api ${{ env.DEVOPS_ASST_API_ARGS }}
+    # - name: Report status
+    #   run: gh api ${{ env.DEVOPS_ASST_API_ARGS }}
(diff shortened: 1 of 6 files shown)


JasperVanDenBosch:
Good catch @JonathanShor !

I'd like us to have a unit test to make sure this won't be re-introduced in the future. If you have a chance to add this, great, otherwise I'll aim to add one later this week.

PS please ignore the style and typing checks, these are not working rn.

JonathanShor:
Sure, I could add something like my example in #452. Are there docs on how your testing is setup? Or point me to an example test and a location I can add the new one, that would be great.

JonathanShor:
Corrected for the case that rdm1 and rdm2 have different n_rdm. Had NaNs propagate across RDMs within rdm1 or rdm2, and it only raises an error if the positions disagree after this propagation.

HeikoSchuett:
I just added a unit test from the issue. 
I am sorry that the auto formatter hit with the unit test file, but this should now be good to go?

HeikoSchuett (approved):

JasperVanDenBosch (approved):

JasperVanDenBosch:
hi guys, sorry this took a while. 
I've undone some of the work of the auto-formatter (so I could see what's going on). We have instructions for formatting in `ruff.toml` but we could add something for other formatters as well.  
The test looks good and passes.
I've also kind of abused this PR to fix some regressions on main, sorry. Not to be taken as an example :P

JasperVanDenBosch:
this is now on pypi as `rsatoolbox==0.3.1.dev26`

PR #457, opened 2025-09-06
Fix cv assignment in calc.py

JonathanShor:
_build_rdms expects this to be the actual fold assignment list, not the descriptor name. Converted it to an NDArray to match the type hint in build_rdm.py.

### DIFF
--- a/src/rsatoolbox/rdm/calc.py
+++ b/src/rsatoolbox/rdm/calc.py
@@ -385,7 +385,7 @@ def calc_rdm_crossnobis(dataset, descriptor, noise=None,
         'crossnobis',
         descriptor,
         noise=noise,
-        cv=cv_descriptor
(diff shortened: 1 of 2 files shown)


JonathanShor:
Hmm, I see that maybe _build_rdms _is_ expecting the cv_descriptor, even tho the type hint wants an array. Should the type hint be for `str` instead?

JasperVanDenBosch:
I think the intention is for this to be a string. I.e. a name of one of the existing descriptors. The type annotation is wrong, it should be `Optional[str]`

JasperVanDenBosch:
Also I can't find anywhere where that `RDMs` attribute is used, so (if any) the purpose would just be provenance I suppose.

JasperVanDenBosch (approved):
Thanks @JonathanShor! Merging with the typing fix

PR #461, opened 2025-10-27
Framed RSA and computing V

johnmarktaylor91:
This pull request is for expanding the functionality for computing V, the variance-covariance matrix of dissimilarity estimates. The reason for this is that the current toolbox does not allow for incorporating the estimates of the true distances between patterns into the calculation of V, but rather only allows for assuming these distances to be zero when estimating V. This is often a practical assumption, but 1) it might be nice to at least give users this option, and 2) in the case of Framed RSA, it ends up being an essential step.

This raises the question of how best to incorporate this into the toolbox. Relevant considerations are as follows:

1) [Heiko has very nicely derived](https://github.com/user-attachments/files/23171663/Cov_of_Crossnobis_RDMs-3.7.pdf) the proper formula for V in the case of the crossnobis distance. This formula is the same as previous formulas for the signal-independent term, but differs in the signal-dependent term. It seems like it would be best to have a single function for computing V, with the option of either including or not including the true distances. I coded up a version where an RDM can be optionally provided for this purpose, along with an optional binary mask (np.array) if the user wishes to incorporate some distances into the estimation of V, but not others (in the case of Framed RSA, this is because there are "frozen" patterns that have no noise and that are of special interest).

2) For completeness, it seems like the function ought to be able to handle euclidean/mahalanobis distance as well, not just crossnobis. It's not obvious to me whether this falls immediately out of the formula for V with crossnobis distance. If it is also   not obvious to others then I will sit down and try to work through the math. 

3) In a separate (and in principle independent) pull request, I've implemented a function for computing sigma_k from a dataset, which doesn't currently exist in the toolbox.

4) We will have to figure out how this new functionality ought to plug into other parts of the toolbox. In particular: currently the `compare` function takes an optional sigma_k argument when whitened RDM comparators are used, and this sigma_k is used to compute V under the hood (invisibly to the user) and whiten the RDMs being compared. This raises the question: if we want to give new options for computing V, how should this be made available to the user? 

One option would be to add new optional arguments to the `compare` function (e.g., for whether or not to incorporate the true distances when estimating V, and an optional mask for specifying which distances). However, this has the potential downside of adding argument bloat to a core function in the toolbox that's only relevant when using whitened RDM comparators. I wonder about the viability of an alternative option: instead of making sigma_k an argument in `compare`, we instead make V an optional argument. This allows for the additional arguments to be contained in a separate `get_v` function (instead of adding bloat to `compare`), while still enabling the same default behavior (i.e., the default option for V would be equivalent to using sigma_k=identity, and not incorporating the true distances). I am open to other options but those are just some initial thoughts.


HeikoSchuett:
I fixed two small bugs that made the tests fail on this pull request. Now things appear to run.
I am sorry about the automatic style fixes I added. This makes the pull request a bit messy to look at now.

In terms of the connection to the other methods:
- for Euclidean and Mahalanobis distances the variances are quite similar (and easier to derive). I am not sure where we have a good written out version of this equation, but we should have one as in Jörns paper.
- using v directly as an input is ok from my point of view. The thing that is important to preserve is the default behaviour and that this default runs faster. If we assume independent measurements of the patterns (sigma_k = identity) the equivalence to CKA gives a much faster solution, which we should keep using! Perhaps this is also something worth checking for your framed RSA stuff.

johnmarktaylor91:
> for Euclidean and Mahalanobis distances the variances are quite similar (and easier to derive). I am not sure where we have a good written out version of this equation, but we should have one as in Jörns paper.

Indeed they end up being almost identical, if I'm not wrong: for crossnobis you divide the noise term by M-1, for euclidean/mahalanobis you divide by M. I'm surprised it ended up being so simple but but simulations seem to check out. I've tweaked the code to account for this. If this seems too good to be true please let me know and I'll re-check the math.

> using v directly as an input is ok from my point of view. The thing that is important to preserve is the default behaviour and that this default runs faster. If we assume independent measurements of the patterns (sigma_k = identity) the equivalence to CKA gives a much faster solution, which we should keep using! Perhaps this is also something worth checking for your framed RSA stuff.

Great suggestions and sounds good. I've tweaked the functions in compare.py accordingly. There are a couple remaining wrinkles: 1) comparing RDMs that use negative riemannian distance still takes sigma_k as an argument but doesn't use it in service of computing V, so this is one area where the new control flow breaks down (any thoughts on how to handle?), and 2) for computing whitened correlation/cosine there is some functionality where you provide a 1-D vector for sigma_k (which I believe yields a diagonal matrix with the specified variances). Under the new setup where the user-facing argument is v rather than sigma_k, it seems like this would have to be removed or relocated to elsewhere. 

One other question: do we want to keep get_v in compare.py, or should we move it to noise.py (or even add it to __init__ somewhere so it can be called without the user having to go too deep down the hierarchy)? As it's now a user-facing function we might want to make it easy to find.

JasperVanDenBosch:
hi both,
Sorry for taking so long to look at this. Thanks again for all your work on this johnmark. 


~I think the best solution for solving the bloat in the signatures and preventing re-calculating things or checking for what was provided is to pass around a an object that encapsulates these related variables and is passed around as, say "cov". The base could be an ABC e.g. "Variance" or  "Covariance" class. Concrete implementations could be e.g. Diagonal / Isotropic / Pattern / Signal.~
~They would expose:~
- ~matrix~
- ~whiten (varies depending on implementation)~
- ~sigma_k~
- ~subsample_patterns?~

~We can cache V on the class (e.g. memoise pattern).~



If this sounds okay I'm happy to take a stab at this for you guys to review. Let me know what you think.

Other than that I think the only solution would be to keep both arguments and do the work of checking. 

As for where get_v goes I'm fine with noise or maybe a new covariance module.

The only other remaining issue seems about coming up with good defaults for n_obs / n_fold in build_rdms for non-cv cases

JasperVanDenBosch:
With some more sleep, I realized that was overengineered and I wonder if we could just calculate / retrieve these variables when we need them. 
- Dataset is responsible for calculating sigma_k
- RDMs receives a reference to sigma_k at calculation.
- RDMs is responsible for calculating V (or retrieving it from cache)
- Let's group the parameters by purpose, so they can be responsible for meaningful defaults etc. See `CovarianceOptions`. Could also be a dictionary if this sounds too intimidating for end user.
- We use caching to prevent re-calculating things
- the base dataset gets a `frame()` method to store FramedRSA relevant things. This could return a thin FramedDataset child class, or just be the same Dataset with additional descriptors. Leaning toward the latter.

```
ds = Dataset()
ds = ds.frame(include_all_zeros, all_c_scale)
cov = CovarianceOptions(
	sigma_k_type = 'patterns', # none/identity/1d/patterns
	other_relevant_setting = 'baz' # e.g. for custom array
) 
rdms = calc_rdm(ds, foo, bar, variance_options) # preserve sigmak
	sigma_k = ds.calc_sigma_k(sigma_k_type, noise)
	rdms.set_sigma_k(sigma_k)
compare(rdms, foo, bar)
	rdms.get_v()
		self.get_sigmak()
```

Also I think for `calc` we should start doing something like:

```
calc_rdm(Minkowski(degree, root))
calc_rdm(Poisson(prior_lambda, prior_weight))
```
i.e. every dissimilarity metric has its own options. 

PS I think it makes sense to merge #464 into this. To prevent the mental gymnastics on figuring out what is necessary for framed RSA in this PR without being able to put it into code.

johnmarktaylor91:
Hi Jasper, many thanks for all your work on this—

- Agree with merging this with #464 for simplicity
- I agree calculating V on-demand and caching is a clean default solution
- For `calc_rdm(Minkowski(degree, root))` I defer to your taste here, this seems like a big change in UI for the toolbox but if you're fine with it I'm fine with it. It could be cleaner than an explosion of arguments that are only applicable to certain dissimilarity metrics
- I like your `frame` method idea, I also lean towards just making it the same dataset with additional descriptors.

JasperVanDenBosch:
Great, I'll start on these and get back to you for review

JasperVanDenBosch:
- [ ] framing interface
- [ ] covariance interface
- [ ] parameterised metrics interface
- [ ] notebook example
- [ ] docs
- [ ] tests
- [ ] caching

johnmarktaylor91 (commented):
All looks good to me!

PR #467, opened 2025-11-20
Fixes #466: Add optional boundary truncation for searchlights

### LINKED ISSUE #466: Including searchlight neighbours outside of mask boundary - is this expected?
In the `util/searchlight.py` the function to return all neighbouring voxels is missing truncation at the boundary:

```python
def _get_searchlight_neighbors(mask, center, radius=3):
    """Return indices for searchlight where distance
        between a voxel and their center < radius (in voxels)

    Args:
        center (index):  point around which to make searchlight sphere

    Returns:
        list: the list of volume indices that respect the
                searchlight radius for the input center.
    """
    center = np.array(center)
    mask_shape = mask.shape
    cx, cy, cz = np.array(center)
    x = np.arange(mask_shape[0])
    y = np.arange(mask_shape[1])
    z = np.arange(mask_shape[2])

    # First mask the obvious points
    # - may actually slow down your calculation depending.
    x = x[abs(x - cx) < radius]
    y = y[abs(y - cy) < radius]
    z = z[abs(z - cz) < radius]

    # Generate grid of points
    X, Y, Z = np.meshgrid(x, y, z)
    data = np.vstack((X.ravel(), Y.ravel(), Z.ravel())).T
    distance = cdist(data, center.reshape(1, -1), 'euclidean').ravel()

    return tuple(data[distance < radius].T.tolist())
```

and later in `get_volume_searchlight` only entire searchlight volumes are dropped based on overlap with the mask, but no truncation is applied.

Unfortunately, when the threshold is <100% the untruncated volumes are going to produce misleading artefacts when comparing against models (potentially the dead ReLUs are having a party here):
Image

A quick fix without touching the code is to just increase the `threshold` parameter to 100% but that removes all the search light centers whose sphere would be truncated.

Question: if the threshold is smaller than 100%, wouldn't it be more pragmatic to still truncate the volumes with the mask?

botcs:
# Add optional boundary truncation for searchlights
Followup PR on #466.

When computing searchlight volumes in `util/searchlight.py` with the `threshold` <100% the untruncated volumes produce misleading artefacts:
image

With this optional argument we can still use <100% thresholding values and not include noise outside of the mask in the RDMs

New usage:
```python
get_volume_searchlight(mask, ..., truncate_at_boundary=True)
```

## Important TODOs
Since the varying size of the searchlight volumes are going to change the statistics there are future TODOs that would fully support this feature. Namely:
- Searchlight Size Tracking through metadata
- Propagate Size Info Through RDMs
  - The weighting scheme requires deciding:
  - How does variance scale with number of voxels?
  - Linear? Square root?
  - Does it depend on the correlation structure? 
- Accordingly modify functions in `inference/evaluate.py` and `inference/result.py`

### DIFF
--- a/src/rsatoolbox/util/searchlight.py
+++ b/src/rsatoolbox/util/searchlight.py
@@ -15,12 +15,17 @@ from rsatoolbox.rdm.calc import calc_rdm
 from rsatoolbox.rdm import RDMs
 
 
-def _get_searchlight_neighbors(mask, center, radius=3):
+def _get_searchlight_neighbors(mask, center, radius=3, truncate_at_boundary=False):
     """Return indices for searchlight where distance
         between a voxel and their center < radius (in voxels)
 
     Args:
         center (index):  point around which to make searchlight sphere
+        radius (int): radius of the searchlight sphere in voxels
+        truncate_at_boundary (bool): if True, only include voxels where mask is True.
+            if False (default), include all voxels within radius regardless of mask value.
+            Setting to True addresses issue #466 by preventing searchlights from including
+            voxels outside the mask boundary.
 
     Returns:
         list: the list of volume indices that respect the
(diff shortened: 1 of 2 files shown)


botcs:
the failing plotting test seem to fail on main too, can you check @JasperVanDenBosch?

JasperVanDenBosch (approved):
Looks good to me

HeikoSchuett:
@botcs thanks for picking up on this! I think we can merge this as is. I am not even sure we would need the backward compatibility, i.e. we may remove this at some future point.

@JasperVanDenBosch  This should probably be taken into account for the future searchlight stuff.

HeikoSchuett (approved):

PR #488, opened 2026-07-13
add a __version__

### LINKED ISSUE #487: No __version__
Seen on conda-forge, but locally with `pip install rsatoolbox` as well:
```
>>> import rsatoolbox
>>> rsatoolbox.__version__
  File "<python-input-1>", line 1, in <module>
    rsatoolbox.__version__
AttributeError: module 'rsatoolbox' has no attribute '__version__'
```
It's pretty standard in scientific python to have this attribute, probably worth adding here

HeikoSchuett:
fixes #487

### DIFF
--- a/src/rsatoolbox/__init__.py
+++ b/src/rsatoolbox/__init__.py
@@ -3,12 +3,8 @@
 """ Top level package: Only imports and organisation
 """
 
-try:
-    from importlib.metadata import version
-
-    __version__ = version("mne")
-except Exception:
-    __version__ = "0.0.0"
+from importlib.metadata import version
+__version__ = version("rsatoolbox")
 
 from . import cengine
 from . import data
--- a/src/rsatoolbox/__init__.py
+++ b/src/rsatoolbox/__init__.py
@@ -3,6 +3,13 @@
 """ Top level package: Only imports and organisation
 """
 
+try:
+    from importlib.metadata import version
+
+    __version__ = version("mne")
+except Exception:
+    __version__ = "0.0.0"
+
 from . import cengine
 from . import data
 from . import inference


HeikoSchuett:
I had a look on how to add a __version__ attribute automatically and found a code block that does create a __version__ attribute the same way MNE-Python does. I am not sure however whether this is the right version we would want to put in here.
Can anyone confirm whether this is the right version tag for us?

src/rsatoolbox/__init__.py
```diff
@@ -3,6 +3,13 @@
 """ Top level package: Only imports and organisation
 """
 
+try:
+    from importlib.metadata import version
+
+    __version__ = version("mne")
```
JasperVanDenBosch: this gets the version for MNE, I think you meant 

`__version__ = version("rsatoolbox")`

src/rsatoolbox/__init__.py
```diff
@@ -3,6 +3,13 @@
...
+except Exception:
+    __version__ = "0.0.0"
```
JasperVanDenBosch: This would silence the exception that might be relevant. What does `0.0.0` mean? I'd suggest removing the try/catch wrapper

src/rsatoolbox/__init__.py
```diff
@@ -3,6 +3,13 @@
 """ Top level package: Only imports and organisation
 """
 
+try:
```
JasperVanDenBosch: Being defined in this location, this code will be run often. I would hope that importlib is highly optimized, but if not there might be a performance cost to this?

JasperVanDenBosch: But I suppose if this approach has been in MNE for a while this might be safe?

JasperVanDenBosch (changes requested):
I haven't used `.__version__` before but if there is a use case I'm ok with this. Some suggestions in the inline comments; key being to change the string to "rsatoolbox".

Thanks

src/rsatoolbox/__init__.py
JasperVanDenBosch: Okay this seems to have been [optimized to a degree ](https://importlib-metadata.readthedocs.io/en/latest/history.html)and should only add a few ms. So ignore this point.

src/rsatoolbox/__init__.py
HeikoSchuett: I am happy to drop the try/catch, this is copy past from elsewhere

